PR #196, opened 2015-01-28
Updated polymerapp template for latest packages

kunaldeo:
Updated polymerapp template for polymer 0.15.5 and paper_elements 0.6.1. The template now uses key-up event because valueChanged was not working with the latest paper_elements package. 

PS: Updated lib/generators/polymerapp_data.dart is not built.


kunaldeo:
Please ignore this pull request. I have created a new pull request (197) with built lib/generators/polymerapp_data.dart using grind build .

PR #197, opened 2015-01-28
Updated polymerapp for latest packages

kunaldeo:
Updated polymerapp template for polymer 0.15.5 and paper_elements 0.6.1. The template now uses key-up event because valueChanged was not working with the latest paper_elements package.

[Review on Reviewable](https://reviewable.io/reviews/google/stagehand/197)


templates/polymerapp/pubspec.yaml
```diff
@@ -8,8 +8,8 @@ environment:
   sdk: '>=1.0.0 <2.0.0'
 dependencies:
   browser: any
-  polymer: '>=0.15.0 <0.16.0'
-  paper_elements: '>=0.5.0 <0.6.0'
+  polymer: '>=0.15.5 <0.16.0'
```
devoncarew: Do we need these new lower limits or can we use `.0`?

templates/polymerapp/lib/main_app.dart
```diff
@@ -2,18 +2,19 @@
 // is governed by a BSD-style license that can be found in the LICENSE file.
 
 import 'package:polymer/polymer.dart';
+import 'dart:html';
```
devoncarew: nit: can you sort the `dart:` imports first?

devoncarew:
Thanks for the PR! Comments in-line.

kunaldeo:
Thanks for the comments. I have got the imports organized. Regarding version numbers, I have updated the paper elements to '>=0.6.1 <0.7.0' . 0.6.1 is the latest release of paper_elements as of now, thats why I have mentioned it. I will check between the releases if the sample breaks :)  .

devoncarew:
It's generally good to keep version constraints wider when possible (unless we are using features or fixes from newer versions). This limits the possibility of creating an unsatisfiable set of version constraints (one dep needs polymer >=0.6.0 < 0.6.2, another polymer >= 0.6.4 < 0.7.0). If we do need polymer `>=0.15.5` and paper_elements >= 0.6.1, sgtm, I just want to make sure that that's the case. Thanks again for the PR!

kunaldeo:
Got your point. After doing extensive testing (sorry for multiple commits) I have found out that , polymer version < 0.15.4 returns warning from ScriptCompactor "private symbols are not supported" and paper_elements  <0.6.1 breaks the sample altogether.

devoncarew:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/stagehand/197)

Reviewed files:
- lib/generators/polymerapp_data.dart @ r3
- templates/polymerapp/lib/main_app.dart @ r2
- templates/polymerapp/lib/main_app.html @ r1
- templates/polymerapp/pubspec.yaml @ r3

---

Great! lgtm

---

kunaldeo:
Thanks a lot for accepting the PR.

PR #202, opened 2015-02-18
Third party templates implementation.

kunaldeo:
Here is the complete working implementation of third-party template system for stagehand. This implementation is based on pub package plugins.
To install a third party package:
```stagehand --git <Git Repo URL>```
You can use this example third party template at  https://github.com/kunaldeo/stagehandtrialtemp. Template has the following structure:
```
├── main.dart
├── pubspec.yaml
└── src
    └── stagehandtrialtemp_data.dart
```
Soon the author of plugins package plans to move main.dart to the bin folder. At the time stagehandtrialtemp repo needs to be updated. Process of producing template files would use the same process stagehand. We can probably add a command to stagehand which compiles a third party template.

### CODE REVIEW

**Comment by sethladd** [comment]
How did we come up with the directory layout for the plugin? I'm guessing we're explicitly NOT trying to make a plugin the same layout as a pub package?

How is the pubspec.yaml used by the plugin?

**Comment by kunaldeo** [comment]
Directory layout is not of a pub package but is of a general dart application (https://github.com/kunaldeo/stagehandtrialtemp).

├── main.dart -- Contains the main plugin code. It passed on a map called pluginData which contains the following:
```
    pluginData['info'] = "Stagehand Trial Template";
    pluginData['description'] = "Stagehnad Trial Template Information";
    pluginData['entrypoint'] = "web/index.html";
    pluginData['data'] = templateData;
```
├── pubspec.yaml -- We used it for following
1. To install the dependencies like ```plugins``` after cloning the pub repository.
2. name: Name field registers itself as the third party template name and it also provides other metadata like version etc. 
└── src
    └── stagehandtrialtemp_data.dart : provides ```list<String> templateData```

sethladd:
Can we use a real class instead of a map for all the plugindata?

kunaldeo:
Limitation of isolates. Plugin uses isolate to send the data back. If the type is missing we will hit segfault on vm. For the plugin structure we will have real classes and everything, just like a regular project. We will have additional option in stagehand will allow us to build the installable plugin. In that step we will generate data files and map. The example template represents the compiled output of a proper project.

sethladd:
Can you rebase this down to a single commit? That makes it easier for us to review and makes for a cleaner merge. Thanks!

kunaldeo:
Sorry about the mess. I have rebased it all under one commit.

sethladd:
Thanks, expect I still see 6 commits for this PR. Maybe we need to open a new PR?

kunaldeo:
Rebased all the commits.

PR #204, opened 2015-02-24
First implementation of third party templates

kunaldeo:
Here is the complete working implementation of third-party template system for stagehand.
To install a third party package:
```stagehand --git <Git Repo URL>```
You can use this example third party template at  https://github.com/kunaldeo/stagehandtrialtemp. Template has the following structure:
```
├── main.dart
├── pubspec.yaml
└── src
    └── stagehandtrialtemp_data.dart
```
This pull request merges all the commits in PR https://github.com/google/stagehand/pull/202 . Please see it for earlier discussions.

### DIFF
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -152,7 +152,7 @@ class CliApp {
         return new Future.error(new ArgError('Template already installed.'));
       }
 
-      logger.stdout("Clonning " + repoName);
+      logger.stdout("Cloning " + repoName);
       io.Process.runSync('git', ['clone', options.rest[0], templateLocation]);
       logger.stdout("Running pub in the template directory");
       io.Process.runSync('pub', ['get'], workingDirectory:templateLocation);
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -122,6 +122,45 @@ class CliApp {
       return analytics.waitForLastPing(timeout: _timeout);
     }
 
+    if (options['git']) {
+      if(options.rest.length == 0) {
+        logger.stderr("No URL specified.\n");
+        _usage(argParser);
+        return new Future.error(new ArgError('no URL specified'));
+      }
+
+      Map<String, String> envVars = io.Platform.environment;
+      String homeDir = envVars['HOME'];
+      String stagehandDataDir = path.join(homeDir, 'stagehandData'); //For storing third party templates
+      
+      String repoUrl = options.rest[0];
+
+      //Get the repo name from URL
+      String expression = r'^((http[s]?|ftp):\/)?\/?([^:\/\s]+)((\/\w+)*\/)([\w\-\.]+[^#?\s]+)(.*)?(#[\w\-]+)?$';
+      RegExp regExp = new RegExp(expression);
+      var matches = regExp.allMatches(repoUrl);
+      var match = matches.elementAt(0);
+      var repoName = match.group(6).replaceFirst(".git", "") //remove trailing git if required
+                     .replaceAll("-", ""); //remove dashes from git repo name if required
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by kunaldeo** [comment]
@sethladd I have merged all the changes into this.

**Comment by sethladd** [comment]
gotcha, thanks!

**Comment by sethladd** [comment]
cc @devoncarew

**Comment by sethladd** [comment]
We should probably update the README with instructions on how to create a plugin, and how to consume a plugin.

**Comment by kunaldeo** [comment]
I am working on a tools/grind.dart file to help create the plugin. Once that is done we can update the README.

File: lib/src/cli_app.dart [hunk 1]
```diff
@@ -122,6 +122,45 @@ class CliApp {
...
+      String templateLocation = path.join(stagehandDataDir, repoName);
+
+      if(io.FileSystemEntity.isDirectorySync(templateLocation)) {
+        logger.stderr("Template already installed. If you want to update the template use the update command\n");
+        _usage(argParser);
+        //TODO: Implement template update command
+        return new Future.error(new ArgError('Template already installed.'));
+      }
+
+      logger.stdout("Clonning " + repoName);
```
andreasolund: Should be "Cloning"

kunaldeo:
Added instructions on how to create and consume a third party template. 

Bootstrap Template: [emptyStagehandTemplate](https://github.com/kunaldeo/emptyStagehandTemplate), This will give developers starting point to build new templates. 
Fully Built Sample Template : [polymerContacts](https://github.com/kunaldeo/polymerContacts)

sethladd:
Thanks very much for the pull request! It spawned a lot of discussions here around the water cooler today, mainly about how best to enable a thriving ecosystem of templates, the differences between declarative and imperative solutions, and maintenance costs.

We're impressed by the isolates-based plugins approach. For systems that require imperative plugins (plugins that have to execute code), it's a solid approach. Isolates give us safety from third-party code, and a way to load the code. The plugin package's API makes it easier to do it.

We asked ourselves: "What do the plugins need to execute?" and we couldn't come up with a solid answer. We know a template needs to ask certain questions and display certain answers, but does it need to execute code? Can we get away with _declarative_ templates?

We then asked ourselves, "If we don't need a plugin to execute code, do we want to support and API for those plugins?" The answer to that was quickly: if we can avoid providing a Dart API for the plugins, that's less for us (and plugin maintainers) to worry about, version, etc.

Ultimately, in order to strive for a simple system that asks very little of our template authors, we couldn't find a really compelling reason to go the plugin/isolates/code approach. A design that uses pub itself to distribute and upgrade templates, and a simple declarative approach to collecting info, seems like the most efficient way to accelerate the growth of an ecosystem of plugins for stagehand.

We are extremely appreciative of your hard work, and we don't think we could have found a solution without your pull request. In the end, we feel we have a simpler way forward that doesn't require isolates, and thus we unfortunately aren't going to merge this PR. Devon will follow up with a design that hopefully satisfies all of our use cases, so we ensure we're allowing template authors to do all the great stuff that you envisioned.

Again, even though we don't plan to merge this, we are extremely appreciative of your time and energy, and want to thank you for being part of this project.

kwalrath:
Here are my notes on the design from talking to Devon.

A template could be in a pub package or a github repo. Say it's a pub package. You’d pub global activate that package. Stagehand would enumerate all the activated packages and look for a file like stagehand.yaml in them. Then it needs to load stagehand packages. That was what stopped them before, but now that the resource API is available (in 1.14), loading stagehand packages should be much easier.

For an example of using the resource API, see [this recent dartdoc change](https://github.com/dart-lang/dartdoc/commit/d5f59376d7cc8c8b97d2b9b312b63667d0265c25).

PR #210, opened 2015-03-13
remove project name validation

### LINKED ISSUE #208: remove the project name validation
We're getting the project name implicitly from the cwd. 

https://github.com/google/stagehand/blob/master/lib/src/cli_app.dart#L161

We still want to keep the project name normalization.

devoncarew:
Fix #208, @pq.

### DIFF
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -158,12 +158,8 @@ class CliApp {
       return new Future.error(new ArgError('project directory not empty'));
     }
 
-    // Validate and normalize the project name.
+    // Normalize the project name.
     String projectName = path.basename(dir.path);
-    if (_validateName(projectName) != null) {
-      logger.stderr(_validateName(projectName));
-      return new Future.error(new ArgError(_validateName(projectName)));
-    }
     projectName = normalizeProjectName(projectName);
 
     if (target == null) {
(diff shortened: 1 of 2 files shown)


pq:
LGTM!

alexander-doroshko:
Should normalizeProjectName() now take care of the first character to match [A-Za-z]?

alexander-doroshko:
@devoncarew @pq Looks like normalizeProjectName() now should take care of the first character to match [A-Za-z]. May be some more normalization needed, for example I'm not sure if Chinese or Russian symbols in the project name will work. And special symbols like `~!@#$%^&*()_+='"\|/<>[]{}

devoncarew:
@alexander-doroshko, agreed! We should do a bit more rigorous normalization. Is this something you'd be interested in contributing? Otherwise I can open a separate issue for it.

alexander-doroshko:
@devoncarew Yes, I can do it myself if you want. Just to make sure: are restrictions for the project name the same as for Dart lang identifier?

devoncarew:
Yup, here are the conventions: https://www.dartlang.org/tools/pub/pubspec.html#name

Some of them are more important than others. I'm not exactly sure what to do if the directory name starts with numbers, or invalid chars. Do we just remove chars until we get to a legal one? Or convert them to more legal chars? Or prefix with a legal char? None of those options are particularly appealing.

PR #216, opened 2015-04-01
fix typo

guillermooo:


lib/src/common.dart
```diff
@@ -15,7 +15,7 @@ import '../stagehand.dart';
 
 const int _RUNE_SPACE = 32;
 
-List<TemplateFile> decodeConcanenatedData(List<String> data) {
+List<TemplateFile> decodeConcatenatedData(List<String> data) {
```
devoncarew: I know this is in a private file, but I'd feel more comfortable if we kept a deprecated version of the method around, that just forwarded to the correct name:
```
@Deprecated('see decodeConcanenatedData')
List<TemplateFile> decodeConcanenatedData(List<String> data) =>
    decodeConcatenatedData(data);
```

devoncarew:
Thanks! One comment. Can you also add yourself to the AUTHORS file in the PR?

guillermooo:
Done!

devoncarew:
Thanks!

PR #225, opened 2015-04-20
automate updating the docs

devoncarew:
@sethladd, push the template descriptions into the documentation as part of a build step, to reduce busy-work when changing templates.


sethladd:
Awesome.

PR #229, opened 2015-04-21
update the appengine template

devoncarew:
- update the appengine template (rev to 0.3.0)
- rename the id from `appengineapp` to `console-appengine`

@wibling, can you take a look at this?

Also, I notice the appengine template has a `build/web/usage.html` file in it. The `build` dir is generally auto-generated - should this template have that file?


devoncarew:
TBR'ing

wibling:
lgtm

Sorry for the delay.

PR #238, opened 2015-04-28
update template descriptions

devoncarew:
- Update the template descriptions
- The `web-full` sample now reverses the text on page load

@sethladd


devoncarew:
(TBR)

PR #246, opened 2015-05-12
define version constraint for unittest in console-full template

kevmoo:
Until we migrate all templates to pkg/test


devoncarew:
lgtm

PR #247, opened 2015-05-12
migrate to pkg/test

kevmoo:


test/mock_test.dart
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2014, Google Inc. Please see the AUTHORS file for details.
+// All rights reserved. Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+library stagehand.web_test;
```
devoncarew: nit: we should update the lib name as well

devoncarew:
lgtm w/ library name change

PR #250, opened 2015-05-13
migrate package-simple and console-full to use pkg/test

kevmoo:
Validate that the tests pass as part of the grinder test task


devoncarew:
awesome, lgtm!

PR #252, opened 2015-05-13
Make the generate validator use the test syntax

kevmoo:
Runs all generators, even if one fails
Updated travis.sh

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -9,7 +9,6 @@ import 'package:ghpages_generator/ghpages_generator.dart' as ghpages;
 import 'package:grinder/grinder.dart';
 import 'package:path/path.dart' as path;
 import 'package:stagehand/stagehand.dart' as stagehand;
-import 'package:yaml/yaml.dart' as yaml;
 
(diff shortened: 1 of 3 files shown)


kevmoo:
@devoncarew will rebase on your last commit

tool/grind.dart
```diff
@@ -59,72 +58,6 @@ void updateGhPages() {
     ..generate();
 }
 
-@Task('Run each generator and analyze the output')
-void test() {
```
devoncarew: Let's keep this task; it can simplify to something like:

    new PubApp.local('test').run(['test/validate_templates.dart']);

kevmoo: Done, although nothing prints until it finishes.

Is there a way to do async tasks in grinder?

kevmoo:
@devoncarew PTAL

tool/grind.dart
devoncarew: There's an issue to track it: https://github.com/google/grinder.dart/issues/195.

devoncarew:
lgtm!

Issue #255, opened 2015-06-05
tweak web-simple


kwalrath:
A couple of things I happened to notice:

* Why do we bother specifying sdk?
  * If we do want to specify sdk, then perhaps it should be `sdk: '>=1.8.3 <2.0.0'`, so we can use caret notation elsewhere (`browser: ^0.10.0`).
* Why does index.html specify `data-pub-inline` in the dart.js script tag? I found a [blog post](http://blog.sethladd.com/2014/04/speed-up-your-dart-apps-initial-load.html) that mentions it, but it seems like it wouldn't work without the script_inliner transformer. It looks like the angularapp template also has this problem. (web-full uses data-pub-inline and script_inliner, so I guess it's fine.)

devoncarew:
Taking out the sdk constraint sounds fine, or bumping it to min. of 1.9.0.

@sethladd for the  `data-pub-inline` question -

sethladd:
For SDK, can we say "<2.0.0" ?

For dart-pub-inline, yes, you need the transformer. Apologies if I forgot to add it to the sample.

sethladd:
Someone (we? me?) needs to add the following to the pubspec.yaml:

```
dependencies:
  script_inliner: ">=0.0.1 <0.1.0"
```

and

```
transformers:
- script_inliner
```

kwalrath:
Should dartlang.org be recommending script_inliner, or are we good with just including it in templates?

Should I be including it in the angular2 examples' pubspecs? Or maybe I could just have a note about it.

Also @sethladd: https://pub.dartlang.org/packages/script_inliner renders the README with ugly, escaped `&lt;` and `&gt;` instead of `<` and `>`. Color me too lazy to figure out where to file that as a bug...

devoncarew:
Or use `dart_to_js_script_rewriter`?

sethladd:
I filed https://github.com/dart-lang/pub-dartlang-dart/issues/31 for the issue with rendering markdown on pub.dartlang.org

sethladd:
I suppose I should update web-full :)

sethladd:
rewriter is now used instead of inliner for both simple and full

PR #258, opened 2015-07-08
use dart_to_js_script_rewriter in web-simple

sethladd:


sethladd:
Addresses https://github.com/google/stagehand/issues/255

devoncarew:
lgtm, I like the `dart_to_js_script_rewriter` solution better then the scrip-inliner one, for remove the browser.js reference.

devoncarew:
Possibly consider a shorter name for the dart to js script rewriter package?

PR #266, opened 2015-09-17
update for polymer 1.0-rc1 (DO NOT MERGE)

jakemac53:
**This version of polymer is not yet released, so please wait to merge**

Fixes https://github.com/dart-lang/polymer-dart/issues/552

### DIFF
--- a/templates/web-polymer/pubspec.yaml
+++ b/templates/web-polymer/pubspec.yaml
@@ -9,9 +9,9 @@ environment:
 
 dependencies:
   browser: ^0.10.0
-  polymer_elements: ^1.0.0-rc1
-  polymer: ^1.0.0-rc1
-  reflectable: 0.2.0
+  polymer_elements: ^1.0.0-rc.1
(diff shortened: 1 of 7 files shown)


devoncarew:
cc @kwalrath, @Sfshaza

templates/web-polymer/pubspec.yaml
```diff
@@ -8,10 +8,20 @@ environment:
   sdk: '>=1.0.0 <2.0.0'
```
donny-dont: Shouldn't the SDK dependency be bumped to `'>=1.9.0 < 2.0.0'` if you want to use the `^` in your dependencies?

Sfshaza: Good catch! Caret syntax was added in 1.8.3, so 1.9.0 seems reasonable.

devoncarew:
lgtm!

jakemac53:
thanks!

Issue #269, opened 2015-10-06
update the polymer dep from 1.0.0-rc.1


devoncarew:
We should push a new version of the template when `1.0.0` is stable, so people's app don't have `1.0.0-rc.1` (or `1.0.0-rc.2`) in them.

@jakemac53

devoncarew:
(new template here: https://github.com/google/stagehand/pull/266)

terrywarwar:
Polymer versions were created with:
polymer_elements: ^1.0.0-rc.2
polymer: ^1.0.0-rc.1

They should be: 
polymer_elements: ^1.0.0-rc.1
polymer: ^1.0.0-rc.2

jakemac53:
doh! I can work on a pull request

devoncarew:
@jakemac53, should we rev to the latest (`-rc5`?)? Or does the current sample work against that one?

jakemac53:
current sample should still work

jakemac53:
In the `rc` phase I am releasing versions pretty regularly (at least once a week) so I don't think its work tracking those for stagehand as long as it works in the latest

devoncarew:
:+1: makes sense to me

PR #272, opened 2015-10-06
Fix up polymer version constraints

jakemac53:
As pointed out in https://github.com/google/stagehand/issues/269 I mixed these up


devoncarew:
lgtm

PR #276, opened 2015-10-16
Kevmoo fixes

kevmoo:
[Review on Reviewable](https://reviewable.io/reviews/google/stagehand/276)


templates/server-appengine/lib/memcache.dart
```diff
@@ -47,8 +46,13 @@ void write(HttpResponse response, Map<String, String> valueMap) {
 /// Helper method to read a set of values from the memcache.
 void read(HttpResponse response, Iterable<String> keys) {
   var memcache = context.services.memcache;
-  Future.forEach(keys, (key) => memcache.get(key)
-      .then((value) => response.writeln('"${key}": "${value}"'))
-      .catchError((_) => response.writeln('"${key}": value not found!')))
-    .whenComplete(response.close);
+  Future
```
devoncarew: The previous version is much more readable. It looks like this was just formatted - no code changes? I wonder if some code changes would produce better formatting. Or perhaps you could terminate lines with ` //` to force the formatting on the left.

tool/grind.dart
```diff
@@ -85,27 +81,22 @@ String _toStr(String s) {
   }
 }
 
-void _traverse(Directory dir, String root, List<String> results) {
+Iterable<String> _traverse(Directory dir, String root) sync* {
```
devoncarew: Interesting - I've never used `sync*` before.

tool/grind.dart
```diff
@@ -60,18 +60,14 @@ test() => new TestRunner().testAsync(files: 'test/validate_templates.dart');
...
-  String str = results.map((s) => '  ${_toStr(s)}').join(',\n');
+  String str = _traverse(src, '').map((s) => '  ${_toStr(s)}').join(',\n');
 
   target.writeAsStringSync("""
 // Copyright (c) 2014, Google Inc. Please see the AUTHORS file for details.
 // All rights reserved. Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-List<String> data = [
+const List<String> data = const <String>[
```
devoncarew: This is pretty explicit typing - what about just `const List<String> data = [ ... ]`? It's my normal pattern. I only use full types on the right when I get some value out of it, like then DDC requests some additional type info in a location.

templates/server-appengine/lib/memcache.dart
kevmoo: This needs to be refactored to use `shelf`. In the mean time, I was pushing for a 100% clean format for all output – it'll make it easier to do future changes.

tool/grind.dart
kevmoo: By interesting, do you mean **freakin' amazing** ?

tool/grind.dart
kevmoo: I'd argue for nuking the type info on the left. Inference does the work for you. The reified generic (on the right) actually has runtime behavior.

templates/server-appengine/lib/memcache.dart
devoncarew: If we're going to auto-format the code here, some re-structuring would make this section more readable.

I think we want to keep these starter templates relatively focused; this one is targeted towards appengine; there is an existing one for users who want to start with shelf.

tool/grind.dart
devoncarew: I use the types on the left as documentation when I'm reading the code - it makes it easier to scan existing code and figure out what's going on. So many good choices about where to put the types though :)

tool/grind.dart
devoncarew: A bit more jaded than that, but this PR is a good learning example of how to use `sync*`.

devoncarew:
Thanks for keeping these up-to-date! lgtm - will land

tool/grind.dart
sethladd: +1 for using language features when they make sense.

tool/grind.dart
sethladd: What do you mean "runtime behavior" ?

templates/server-appengine/lib/memcache.dart
sethladd: I don't think anyone should be using app engine without shelf. It's OK to combine them.

tool/grind.dart
kevmoo:
```dart
List<int> foo = [1, 2, 3];
assert(foo is List<int>); // fails
```

```dart
var foo = <int>[1,2,3];
assert(foo is List<int>); // succeeds
```

The generic definition on the right affects runtime behavior. The one on the left does not.

sethladd: Right. But do you actually do that in your code here?

kevmoo: Nope. The other case is adding an item to the list, but this is a const, so that's not an issue either.

But for both cases, I'd argue for putting types on the right all the time – good to model it for users.

sethladd: Not convinced, but let's take this offline and discuss. Certainly don't
want to hold up improvements to Stagehand over a tiny thing like this.
Also, thanks for the patches!

PR #280, opened 2015-10-17
Migrate server-appengine sample to use pkg/shelf

kevmoo:
[Review on Reviewable](https://reviewable.io/reviews/google/stagehand/280)


devoncarew:
As per the conversation here, I think we should keep the appengine sample simple. We already have a sample to showcase shelf.

@wibling is the original author; he or @sethladd may have other ideas -

kevmoo:
@devoncarew Actually, I think this *is* simple. The low-level API is good – just to have the bare metal exposed – but it's not very easy to understand if you're writing servers.

I think we should promote shelf – because it's easier to understand, compose, etc.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## 0.2.4
-- Improved layout of `server-appengine`
+- Migrated `server-appengine` to use `shelf`.
```
devoncarew: nit: the other changelog entries in this file are just sentence fragments - they don't end in periods.

templates/server-appengine/pubspec.yaml
```diff
@@ -9,3 +9,5 @@ environment:
 
 dependencies:
   appengine: '>=0.3.0 <0.4.0'
+  shelf: ^0.6.0
```
devoncarew: perhaps change the dep above to `^0.3.0` for uniformity with the others?

devoncarew:
lgtm - it looks like we're good to go for adding shelf to this.

wibling:
LGTM. I am fine with promoting shelf instead of the standard dart:io.

Andersmholmgren:
\o/

kevmoo:
@devoncarew Crap. Missed the nits. Will send another PR

PR #281, opened 2015-10-22
Update gitignore templates

ArgentiApparatus:
1. Include pattern for `pubspec.lock` in package-simple library package
only
1. Remove pattern for IntelliJ specific `.idea`
1. Remove pattern for MacOS specific `.DS_Store`
1. Update patterns for files and directories generated by *pub*
1. Add patterns for files generated by *dart2js*
1. Add pattern for directory created by *dartdoc*

Patterns for *pub*, *dart2js* and*dartdoc* are designed to work with Dart SDK 1.12, and are in sync with Dart .gitignore at https://github.com/github/gitignore.

Sources:

* https://www.dartlang.org/tools/private-files.html
* https://stackoverflow.com/questions/20314796/which-files-are-generated-when-executing-dart2js-and-why
* https://www.dartlang.org/tools/dart2js/

### DIFF
--- a/templates/console-full/.gitignore
+++ b/templates/console-full/.gitignore
@@ -7,11 +7,17 @@ build/
 **/packages/
 
 # Files created by dart2js
-*.js
-*.precompiled.js
+*.dart.js
+*.part.js
 *.js.deps
 *.js.map
 *.info.json
 
 # Directory created by dartdoc
-doc/
+doc/api/
+
+# JetBrains IDEs (WebStorm and IDEA are the recomended Dart IDEs) 
+.idea/
+*.iml
+*.ipr
+*.iws
--- a/templates/console-full/.gitignore
+++ b/templates/console-full/.gitignore
@@ -2,8 +2,8 @@
 .buildlog
 .packages
 .project
-.pub
-build
+.pub/
+build/
 packages
 pubspec.lock # (Remove this pattern if you wish to check in your lock file)
 
--- a/templates/console-full/.gitignore
+++ b/templates/console-full/.gitignore
@@ -2,9 +2,10 @@
 .buildlog
 .packages
 .project
-.pub/
-build/
-**/packages/
+.pub
+build
+packages
+pubspec.lock # (Remove this pattern if you wish to check in your lock file)
 
 # Files created by dart2js
 *.dart.js
--- a/templates/console-full/.gitignore
+++ b/templates/console-full/.gitignore
@@ -1,8 +1,17 @@
+# Files and directories created by pub
 .buildlog
-.DS_Store
-.idea
 .packages
+.project
 .pub/
 build/
-packages
-pubspec.lock
+**/packages/
+
+# Files created by dart2js
+*.js
+*.precompiled.js
+*.js.deps
+*.js.map
+*.info.json
+
+# Directory created by dartdoc
+doc/
--- a/templates/package-simple/.gitignore
+++ b/templates/package-simple/.gitignore
@@ -7,14 +7,20 @@ build/
 **/packages/
 
 # Files created by dart2js
-*.js
-*.precompiled.js
+*.dart.js
+*.part.js
 *.js.deps
 *.js.map
 *.info.json
 
 # Directory created by dartdoc
-doc/
+doc/api/
 
-# This is a library package
+# This is a library package, so ignore pub lock file
 pubspec.lock
+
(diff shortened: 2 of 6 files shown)


sethladd:
Why _not_ ignore .idea and .DS_Store?

ArgentiApparatus:
My thinking was that the users needs to choose additional patterns based on their development setup.

I can see the case for leaving the .DS_Store pattern - I'm guessing that is fairly standard pattern in gitignore for MacOS developers.

If IntelliJ specific patterns are included, shouldn't patterns for other common Dart development editors and IDEs be included too? (Sublime Text 3, Atom, Eclipse, Emacs and Vim.)

Maybe just the intelliJ pattern should be provided (as it's the recommended IDE), but a comment be left for the user to guide them to https://github.com/github/gitignore to find patterns for the tools they intend to use.

templates/console-full/.gitignore
```diff
@@ -1,8 +1,17 @@
...
 .buildlog
-.DS_Store
-.idea
 .packages
+.project
 .pub/
 build/
-packages
-pubspec.lock
+**/packages/
```
devoncarew: `packages` works - it matches the symlinks (currently) generated by pub.

templates/console-full/.gitignore
```diff
@@ -1,8 +1,17 @@
+# Files and directories created by pub
 .buildlog
-.DS_Store
-.idea
 .packages
+.project
 .pub/
 build/
-packages
-pubspec.lock
```
devoncarew: I hardly ever check in `pubspec.lock` files - just in cases where there's a hard requirement that developers / build systems / deployment are all on the exact same versions.

templates/console-full/.gitignore
ArgentiApparatus: True, I had made patterns for directories explicitly so (`pub/`, `build/` ...) so `**/` is required on front of `**/packages/`. My thinking was 'let's make it more obvious what these patterns are supposed to match with'.

pub patterns might equally well be:
```
.buildlog
.packages
.project
.pub
build
packages
```

templates/console-full/.gitignore
ArgentiApparatus: Me neither, but https://www.dartlang.org/tools/private-files.html says to check it in for application pacakges. Stagehand being out of sync with 'official' Dart recommendations would be confusing for beginners.

Adding a comment may be useful.

In package-simple:

```
# This is a library package, so ignore pub lock file
pubspec.lock
```

in others:

```
# Usual practice is to commit the pub lock file in application packages
# Un-comment the following line if you wish to ignore it
# pubspec.lock
```

ArgentiApparatus:
My branch is updated:
* Add Jetbrains patterns back in
* Updated dart2js patterns based on https://github.com/dart-lang/www.dartlang.org/issues/1496
* doc/ → doc/api/

sethladd:
Thanks! lgtm

sethladd:
@devoncarew any thoughts?

devoncarew:
I'd like to keep `packages` as is, and would like the lock file ignored for most templates.

ArgentiApparatus:
I'm a bit concerned that always ignoring the lock file goes against https://www.dartlang.org/tools/private-files.html. Although I'm not going to argue over it a great deal.

Would a comment in the templates (other than the simple-package template) be acceptable?

```gitignore
# Ignore pub lock file (remove this pattern if you wish to check in your lock file)
pubspec.lock
```

ArgentiApparatus:
Another commit pushed

devoncarew:
+1, I like having the comment in there.

My reason for having a bare `packages` is that that will match the top-level packages dir, which is a directory, and the packages in the child folders, which are symlinks, and would not be matched by a folder glob. So it matches `packages/` as well as `web/packages`.

Keeping the trailing slash on things like `.pub/` and `build/` totally makes sense however. Happy to land this as is, and I'll put trailing slashes back on the `.pub/` and `build/`, or you can update. Thanks for the PR!

ArgentiApparatus:
A leading`**` followed by a slash means match in all directories. `**/packages/` matches

    packages/
    example/packages/
    test/packages/

etc. in my projects on (Windows 7). If bare `packages` is preferred that's fine by me.

ArgentiApparatus:
Commit pushed

devoncarew:
re: 

> A leading** followed by a slash means match in all directories. **/packages/ matches

Good to know! Sounds like this or `packages` would work fine. I think I'm just used to the current. In any case, thanks again for the PR!

PR #282, opened 2015-10-31
Update index.html

Medeah:


kevmoo:
We generally lean towards 2-space indents, I think.

Also: you should re-run the generator to update the checked-in code that maps to this file.

kevmoo:
@devoncarew thoughts on this?

devoncarew:
@Medeah, thanks for the PR, but I think we'll hold steady with the 2-space indent :)

PR #284, opened 2015-11-05
Update reflectable entry point format

jakemac53:
If people use html imports containing dart scripts then the reflectable transformer needs to run on the bootstrap file, otherwise it can't reach the entire program.

### DIFF
--- a/templates/web-polymer/pubspec.yaml
+++ b/templates/web-polymer/pubspec.yaml
@@ -18,7 +18,7 @@ transformers:
 - web_components:
     entry_points: web/index.html
 - reflectable:
-    entry_points: web/index.dart
+    entry_points: web/index.bootstrap.initialize.dart
 - $dart2js:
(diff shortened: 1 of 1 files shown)


devoncarew:
lgtm

We'll need to re-gen the template resources and rev the pubspec version to re-publish.

jakemac53:
ok, is that something you want me to do?

devoncarew:
That would be great! If I recall correctly, `grind build` should re-build the resources from the template/ dir contents.

jakemac53:
ok I did that, but it looks like it affected a lot more than just the polymer stuff....

devoncarew:
Odd; are you synced up to head? We had a recent PR that adjusted the gitignore template files.

devoncarew:
Otherwise we can just land your first commit.

jakemac53:
Ok, I reverted it

devoncarew:
landing :)

PR #287, opened 2015-11-26
Updates from kevmoo

kevmoo:


lib/src/common.dart
```diff
@@ -52,9 +52,9 @@ String normalizeProjectName(String name) {
 
 /**
  * Given a String [str] with mustache templates, and a [Map] of String key /
- * value pairs, substitute all instances of `{{key}}` for `value`. I.e.,
+ * value pairs, substitute all instances of `A__key__Z` for `value`. I.e.,
  *
- *     Foo {{projectName}} baz.
+ *     Foo A__projectName__Z baz.
```
devoncarew: +1 to making the files valid from the pov of yaml and dart. What about just bracketing the variable with double underscores? It's a bit more readable than `A__foo__Z` and should still be valid in all the same places.

lib/src/common.dart
```diff
@@ -83,9 +83,7 @@ String convertToYamlMultiLine(String str) {
 /**
  * Break the given String into lines wrapped on a [col] boundary.
  */
-List<String> wrap(String str, [int col = 80]) {
-  List<String> lines = [];
-
+Iterable<String> wrap(String str, [int col = 80]) sync* {
```
devoncarew: The previous version is simpler.

test/mock_test.dart
```diff
@@ -24,22 +25,20 @@ Future _testGenerator(Generator generator) async {
   await generator.generate('foo', target);
 
   // Run some basic validation on the generated results.
-  expect(target.getFileContentsAsString('.gitignore'), isNotNull);
-  expect(target.getFileContentsAsString('pubspec.yaml'), isNotNull);
+  expect(target.getFileContentsAsString('pubspec.yaml'), isNotEmpty);
```
devoncarew: Do not all the samples have .gitignores?

devoncarew:
Thanks for the PR! I like the change to make the files analyzable / valid. You've bunched several unrelated changes into one PR however; I would appreciate them as separate PRs.

lib/src/common.dart
kevmoo: If you're not used to `sync *`, sure.

It's a great convention to get into. It can be a big runtime win to avoid allocating and resizing `List`.

lib/src/common.dart
kevmoo: Done!

kevmoo:
@devoncarew PTAL

I keep each commit clean and to-the-point – since it's easy to keep individual commits on GitHub (unlike Rietveld) are you okay keeping these together?

devoncarew:
> I keep each commit clean and to-the-point – since it's easy to keep individual commits on GitHub 

Yup, you have great commit discipline - very semantic commits. It's more that you bunch several unrelated types of changes into one PR. It makes reviewing it tough. 

I'll take a look at this tomorrow.

devoncarew:
@kevmoo, can you:

- open an PR with the fix to `mock_test.dart`
- open an new PR with the `{{ ... }}` ==> `__ ... __` change

Once those are in, a new angular template can use the underscore templating instead of the mustache one.

kevmoo:
@devoncarew done

PR #299, opened 2016-01-22
Update polymer to latest rc.

ntaoo:
From rc.2 to rc.15.


devoncarew:
cc @jakemac53

templates/web-polymer/pubspec.yaml
```diff
@@ -9,18 +9,13 @@ environment:
 
 dependencies:
   browser: ^0.10.0
-  polymer_elements: ^1.0.0-rc.1
-  polymer: ^1.0.0-rc.2
-  reflectable: ^0.3.1
-  web_components: ^0.12.0
```
jakemac53: You do actually need the web_components dependency still (for the polyfills)

templates/web-polymer/web/main.dart
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) __year__, __author__. All rights reserved. Use of this source code
 // is governed by a BSD-style license that can be found in the LICENSE file.
-library my_project.web.index;
+library my_project.web.main;
```
jakemac53: We had intentionally chosen `index.dart` over `main.dart`, its generally simpler if the html/dart file names match up, especially if you have more than one entry point. Not a huge deal but my preference is still for `index.dart`.

jakemac53:
Some minor notes, otherwise lgtm.

templates/web-polymer/web/main.dart
ntaoo: Okey, I will revert this change. Thanks for sharing the reason.

templates/web-polymer/pubspec.yaml
ntaoo: Polymer has web_components dependency. So I believe although deleting web_components dependency here, `index.html` can still have reference to `packages/web_components/webcomponents-lite.min.js`.
But I guess retaining the dependency here is better for the style and future change?

jakemac53: Yes, you don't want to rely on transitive dependencies because removing that dep on the polymer side would result in your app breaking even though it wasn't a breaking change.

devoncarew:
merging...

ntaoo:
Thanks!

PR #302, opened 2016-02-05
Update CHANGELOG.

kwalrath:
Any objections to making the next release 1.0.0?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,21 +1,22 @@
 # CHANGELOG
 
+## 1.0.0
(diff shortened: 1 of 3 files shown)


sethladd:
Nope!

lgtm

devoncarew:
> Any objections to making the next release 1.0.0?

Only that we're not reving it further!

You'll need to update the pubspec version and the constant in `lib/src/cli_app.dart`. Then lgtm2

kwalrath:
OK, I've updated those 2 files.

PR #303, opened 2016-02-05
async -> defer

### LINKED ISSUE #286: Change async modifier to defer in Dart scripts
An async script executes before the whole page is available (in chrome), but defer waits until the page has loaded. In cases where you want to make sure that all the DOM on the initial page is available from within your main function, you should use defer instead of async.

```<script defer src="index.dart" type="application/dart"></script>```
```<script defer src="packages/browser/dart.js"></script>```

Pull requests welcome.

kwalrath:
closes #286


devoncarew:
You may want to update the changelog as well to mention this.

kwalrath:
Already done. Directly in GitHub, sneaky as that seems.

PR #304, opened 2016-02-08
Use git user name for author name

sh4869:
If working directory is git repository and user doesn't set `--author` option, Use git user name for author name.

### DIFF
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -167,8 +167,25 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
-    if (author == '<your name>'  && io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
-      author = io.Process.runSync('git', ['config', 'user.name']).stdout.trim();
+    if (author == '<your name>' &&
+        io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
+      _out(
+          "This directory is git repository.  Try to use git user name for author name ...");
+      try {
+        var stdout = io.Process.runSync('git', ['config', 'user.name']).stdout;
+        if (stdout != null) {
+          author = stdout.trim();
+          _out('Use Git user name for author name.');
+        } else {
+          _out('Error: git user.name is not set.');
+        }
+      } catch (e, st) {
+        if (e is io.ProcessException) {
+          _out('Error: ${e.executable} is not on your path.');
+        } else {
+          return new Future.error(e, st);
+        }
+      }
     }
 
     Map vars = {'author': author};
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -167,6 +167,10 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
+    if (author == '<your name>'  && io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
+      author = io.Process.runSync('git', ['config', 'user.name']).stdout.trim();
+    }
+
     Map vars = {'author': author};
 
     Future f = generator.generate(projectName, target, additionalVars: vars);
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -167,8 +167,7 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
-    if (!options.wasParsed('author') &&
-        io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
+    if (!options.wasParsed('author')) {
       try {
         io.ProcessResult result =
             io.Process.runSync('git', ['config', 'user.name']);
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -177,11 +177,11 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
           author = stdout.trim();
           _out('Use Git user name for author name.');
         } else {
-          _out('Error: git user.name is not set.');
+          _out('Error: git user name is not set.');
         }
       } catch (e, st) {
         if (e is io.ProcessException) {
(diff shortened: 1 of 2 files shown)


lib/src/cli_app.dart
```diff
@@ -169,6 +169,10 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
+    if (author == '<your name>'  && io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
+      author = io.Process.runSync('git', ['config', 'user.name']).stdout.trim();
```
sethladd: What if this fails?

sethladd:
I like the idea, thanks for the patch. Left some comments.

sh4869:
I think if this fails, this should only outputs error info and don't exit because this is not main function of this tool. I write some code following this policy, please check.  Sorry for my poor English.

sh4869:
cc @sethladd

lib/src/cli_app.dart
```diff
@@ -169,6 +169,27 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
...
+        var stdout = io.Process.runSync('git', ['config', 'user.name']).stdout;
+        if (stdout != null) {
+          author = stdout.trim();
+          _out('Use Git user name for author name.');
+        } else {
+          _out('Error: git user name is not set.');
+        }
+      } catch (e, st) {
+        if (e is io.ProcessException) {
+          _out('Error: git is not on your path or fail to run.');
```
sethladd: "or failed to run"

lib/src/cli_app.dart
```diff
@@ -169,6 +169,27 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
+    if (author == '<your name>' &&
+        io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
```
sethladd: I think this needs to be unindented.

lib/src/cli_app.dart
```diff
@@ -169,6 +169,27 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
...
+      _out(
+          "This directory is git repository.  Try to use git user name for author name ...");
```
sethladd: Let's use user.name instead of "user name"

lib/src/cli_app.dart
```diff
@@ -169,6 +169,27 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
...
+      try {
+        var stdout = io.Process.runSync('git', ['config', 'user.name']).stdout;
+        if (stdout != null) {
+          author = stdout.trim();
+          _out('Use Git user name for author name.');
+        } else {
+          _out('Error: git user name is not set.');
```
sethladd: please use user.name instead of "user name"

lib/src/cli_app.dart
```diff
@@ -169,6 +169,27 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
...
+        }
+      } catch (e, st) {
```
sethladd: please use full variable names

kwalrath:
@sethladd should I merge this?

sethladd:
I haven't reviewed the code, but I like the intent of the patch. :)

kwalrath:
@devoncarew can you review the code?

devoncarew:
Hey @sh4869, I know you've had a lot of review on this small patch, so I apologize in advance for the last-minute review. Here are my thoughts:

- we don't want to display any errors to the user about being unable to run `git config`; it's probably not something they're directory expecting stagehand to do
- we want to be a bit more resilient if git returns an error code

I would suggest something like this:

```
if (!options.wasParsed('author')) {
  ProcessResult result = io.Process.runSync('git', ['config', 'user.name']);
  if (result.exitCode == 0) {
    author = result.stdout.trim();
  }
}
```

kwalrath:
Also, @sh4869 could you please add yourself to the AUTHORS file?

sh4869:
I fixed code and added name to the AUTHORS.

lib/src/cli_app.dart
```diff
@@ -169,6 +169,14 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
+    if (!options.wasParsed('author') &&
+        io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
+      try {
+        var stdout = io.Process.runSync('git', ['config', 'user.name']).stdout;
```
devoncarew: This doesn't check the return code - `git config` could return a non-zero code, in which case we don't want to use it's output.

lib/src/cli_app.dart
```diff
@@ -169,6 +169,14 @@ additional analytics to help us improve Stagehand [y/yes/no]? """);
 
     String author = options['author'];
 
+    if (!options.wasParsed('author') &&
+        io.FileSystemEntity.isDirectorySync(dir.path + "/.git")) {
```
devoncarew: I'm not sure if this check is helpful. The `dir` directory will have just been created. We just want to try shelling out to git if the user didn't supply an author option.

sh4869: I fixed it.

sh4869:
ping

lib/src/cli_app.dart
devoncarew: Can you remove this portion of the check: `io.FileSystemEntity.isDirectorySync(dir.path + "/.git")`? We don't need to be conditional on the presence of this directory.

kwalrath:
Sorry, @sh4869, I was out of town. @devoncarew, do you approve this PR?

devoncarew:
Yes, lgtm. Thanks!

kwalrath:
@sh4869 if you can resolve the conflicts, I'll merge this. Thank you!

sh4869:
@kwalrath I did `git rebase`.

kwalrath:
Thanks, @sh4869!

PR #307, opened 2016-02-12
remove library keyword

Kleak:


templates/console-full/lib/__projectName__.dart
```diff
@@ -2,7 +2,6 @@
 // is governed by a BSD-style license that can be found in the LICENSE file.
 
 /// The __projectName__ library.
```
kwalrath: We should delete this comment.

templates/package-simple/lib/__projectName__.dart
```diff
@@ -4,7 +4,6 @@
 /// The __projectName__ library.
```
kwalrath: We should test whether dartdoc treats library comments without a library directive correctly yet...

See https://github.com/dart-lang/dartdoc/issues/1082.

templates/server-appengine/lib/memcache.dart
```diff
@@ -45,7 +43,7 @@ Future write(Map<String, String> valueMap, StringSink buffer) async {
 }
 
 /// Helper method to read a set of values from the memcache.
-Future read(Iterable<String> keys, StringSink buffer ) async {
+Future read(Iterable<String> keys, StringSink buffer) async {
```
kwalrath: Good catch!

kwalrath:
Good stuff, thanks for doing this! We should check whether library docs are generated correctly before committing anything that demonstrates providing library doc comments.

kwalrath:
Oh, and... this PR is related to issue #307.

Kleak:
we have two time this comment :
```dart
/// The __projectName__ library.
```
so i remove both ?

kwalrath:
Yep, remove both please.

kwalrath:
Unless, of course, generating docs for libs doesn't work without the library. In which case, you should restore the command AND the `library` line.

Could you test what dartdoc does?

Kleak:
I will test tomorrow, here it's 1am so i will sleep

Kleak:
it's not working correctly without the library keyword for the library documentation

the incorrect screen : 
screenshot 2016-02-13 at 11 21 53 am

and the correct :
screenshot 2016-02-13 at 11 22 51 am

Kleak:
so i will re-write the ```dart library __projectName__; ``` in two template :
- console-full
- package-simple

devoncarew:
As of recently, dartdoc will take the docs from the first library directive - either the `library` one, or if the file doesn't have that, the first `import` or `export`. We'll probably need to keep the `library` directive for files with dartdoc but without `import`s.

Kleak:
So for this two special case i put ```dart library __projectName__;``` ?
Or we can remove it and the use will complete the file with and import or export ?

Kleak:
On this code : 
```dart
/// The __projectName__ library.
///
/// This is an awesome library. More dartdocs go here.
library __projectName__;

export 'src/__projectName___base.dart';

// TODO: Export any libraries intended for clients of this package.
```

if i remove the library line all the documentation comment is not in the generated documentation

devoncarew:
Make sense to me to keep the library directive when we need to have library docs; sounds like from her comment above that @kwalrath thinks so as well -

Kleak:
with this one it should be ok :)

Kleak:
Ping ?

kwalrath:
LGTM

templates/package-simple/lib/__projectName__.dart
```diff
@@ -1,11 +1,9 @@
 // Copyright (c) __year__, __author__. All rights reserved. Use of this source code
 // is governed by a BSD-style license that can be found in the LICENSE file.
 
-/// The __projectName__ library.
-///
 /// This is an awesome library. More dartdocs go here.
```
kwalrath: This comment should follow good library doc comment practices, including a single noun phrase as the first paragraph. So maybe:

/// Support for doing something awesome.
///
/// This library provides ....

kwalrath:
I just noticed one little problem with the library doc comment. I can fix this myself, unless you really want to do it.

kwalrath:
One last thing, if you wouldn't mind... Could you add yourself to the [AUTHORS](https://github.com/google/stagehand/blob/master/AUTHORS) file? It doesn't have to be in this PR if a separate one would be easier.

Kleak:
done

kwalrath:
Thanks!

PR #314, opened 2016-03-16
Fix the web-polymer template

### LINKED ISSUE #313: web-polymer template doesn't work
A new **web-polymer** project doesn't run as expected: the project is generated with no errors, the dependencies get updated correctly via pub get, but when you run the project with pub serve in Dartium or Chrome, the page is blank. 
Inspecting the DOM you can see an empty `<main-app></main-app>` tag. So I think polymer is not being initialized properly. 

It only happens with version **1.0.2** of stagehand. If I activate the previous version (1.0.1) the generated project works. I have reproduced the error in two Windows machines (WIndows 7 64 bits and WIndows 10 64 bits). 

**Steps to reproduce the problem**: 
(Activate the new version of stagehand)
`pub global activate stagehand`
(Generate the new project)
```
mkdir new_project
cd new_project
stagehand web-polymer
```
(Update dependencies)
`pub get`
(Serve the project)
`pub serve`
(Run in dartium or chrome localhost:8080 and the page is blank). 

**Workaround:**
Install a previous version of stagehand
```
pub global deactivate stagehand
pub global activate stagehand 1.0.1
mkdir new_project_2
cd new_project_2
stagehand web-polymer
pub get
pub serve
```
(The sample code works!)

The same problem happens when I create the project from WebStorm. I suppose it is using the latest version of stagehand internally.

kwalrath:
closes #313

@jakemac53 for review


jakemac53:
Lgtm

kwalrath:
Thanks!

PR #315, opened 2016-03-28
Update web-angular template & dart_to_js_script_rewriter version

kwalrath:


kwalrath:
@kegluneq could you review, please?

kegluneq:
LGTM!

kwalrath:
Per @devoncarew, we're ignoring the coverage/coveralls message.

> Those templates are already covered during the build process - it creates each sample, runs pub on it, and analyzes the result. We could dial back the aggressiveness of that coverage service a bit.

PR #316, opened 2016-03-29
Fix example on stagehand.pub

goderbauer:
`stagehand webapp` results in an error on my system. Changed to `stagehand web-simple`, which works.


kwalrath:
Thanks for the PR!

PR #322, opened 2016-05-23
Changed shelf listener address

Scorpiion:
Hi,

In order to make it easier to deploy simple test apps created with stagehand I suggest that the listener address is changed so that it works both on localhost and on any hosted platform, standalone VMs or PaaS alternatives. As a reference the [app engine example](https://github.com/dart-lang/appengine/blob/master/lib/src/server/server.dart#L30) already uses '0.0.0.0' instead of localhost.

### DIFF
--- a/templates/server-shelf/bin/server.dart
+++ b/templates/server-shelf/bin/server.dart
@@ -22,7 +22,7 @@ void main(List<String> args) {
       .addMiddleware(shelf.logRequests())
       .addHandler(_echoRequest);
 
-  io.serve(handler, 'localhost', port).then((server) {
+  io.serve(handler, '0.0.0.0', port).then((server) {
     print('Serving at http://${server.address.host}:${server.port}');
   });
 }


Scorpiion:
I signed it!

kwalrath:
Thanks, @Scorpiion!

@devoncarew or @kevmoo, any reason not to accept this change? I will, of course, regen.

kevmoo:
Hrm...I get it. We tend to use localhost for our tools – but obviously if you *want* to make a server, you likely want folks beyond the current machine to do it.

BUT: if you're deploying the app, there are many cases where you don't want to bind to any – you likely want to bind to a specific interface.

I think it would be better to add an argument to specify the host – with the default being 'localhost'.

It aligns well with `port`. Usually you want to listen on `80` – but not when developing.

Scorpiion:
Hi, while there sure are cases where you want to bind to a specific ip, I would argue that it is a far more common case that you only have a single routable ip address and that's the one you want to bind to. Especially for a simple getting started template like this. I think it's fair to say that the majority of deployments today are either made to a VM with a single NIC or a container, in both cases there is only one address. 

Comparing the host and port is not really fair here, you have to specify a port, you don't have to specify an ip address. For most deployments you either listen to "0.0.0.0" or hard code the ip of the server/container, where the second alternative don't give you any benefits (unless you have multiple interfaces/ip and the service is an internal only service, as I said before that is not the common case). Pretty much every container and most VMs on all the cloud platforms have a single interface. 

Port 80 would be a bad choice because it's a privileged port, forcing the user to start the server as root, that is not good. Port 8080 is an assign HTTP Alternate port for purposes like this server. 

In a theoretical scenario where someone is deploying this sample to a server with dual nic/ip's, the server in question should have a firewall either way to protect the host. Going further it would also be a rare case where you have a dual nic server and you want to expose a web server only to the internal nic and not to the external one, if you want to expose it to the external one and now "by accident" expose it internally, well that don't make any difference since the external nic already is accessible from the internal network. I think it's an unnecessary mandatory step, it pretty much a guarantee that most people who try to deploy this template will fail on their first attempt. 

I do get the security angle, but it's a weak angle in this case, I simply think this is one of those cases where usability is more important. The server in question is not secure either way, it has no authentication and no encryption, but that not the point, the point is to provide a simple getting started example. At the end of the day it still up to the user to secure their server and listening to localhost here will not make them any safer. 

_I personally don't use Google's docker images, but I guess it's worth mentioning also that the [Dart runtime](https://hub.docker.com/r/google/dart-runtime/) image expects your server app to listen to port 8080, and the [Dart Hello](https://github.com/dart-lang/dart_docker/blob/master/hello/bin/server.dart) image bind to address "0.0.0.0" and port 8080._

Scorpiion:
What's your opinion @devoncarew?

devoncarew:
Thanks @Scorpiion :) I'll largely defer to @kevmoo here, as the person who provided the content for the shelf example. It's a small, 30 line app template, so whatever keeps it simplest and clearest is best imho.

Scorpiion:
Check, it's up to @kevmoo then, it's not the end of the world either way for me, but I think I've made my case. Hardcoding the address to "0.0.0.0" instead of "localhost" would makes things simpler and improve the developer experience I think.

kevmoo:
Chatted a bit with @kwalrath – sorry for the lag here. Thanks for the contribution!

Scorpiion:
No problem, happy to see it merged! :+1:

kwalrath:
I've released stagehand, so this change should be visible everywhere now. Thanks, @Scorpiion and @kevmoo!

PR #330, opened 2016-08-09
Fix comment in gitignore templates

ben4ever:
gitignore files don't allow inline comments. All comments have to be on a separate line.
This leads to the issue that the `pubspec.lock` line in the gitignore templates isn't working as expected.


ben4ever:
I now also updated the generator compressed files and added myself as an author (after reading the wiki, sry :disappointed: ).

kevmoo:
Thanks! Closed in 822325631378f231c4ad5c0e56a06467b8a3357c

Issue #332, opened 2016-08-18
Create app_component.css


filiph:
The angular2 template currently doesn't have app_component.css (and therefore no styleUrls record). I think most people will want to have styling in their components sooner or later (probably sooner).

It's just one additional line in app_component.dart and it's our chance to educate-by-example about:

* how one does styling in AngularDart
* `:host`

If there's a consensus, I'll prepare the PR.

chalin:
Adding a style file for the component is a good idea, but I'm not sure that putting `:host` in it makes the most sense. Why not use `body` instead?

cc @kwalrath

filiph:
I'm not knowledgeable enough about CSS (in or outside of Angular 2) to really have a strong opinion on this.

Wouldn't a `body` selector leak out of the component? Would it even apply (the component doesn't contain a `body` element)? I'm not able to check at the moment. In general, we want to use something that people can just copy-paste to other components and things will just work.

I see a lot of `:host` in component CSS. Actually, looking at Google's AngularDart components (unreleased as of today), it seems to be the _only_ way used to style the whole component.

chalin:
Right, I'm not even sure what the behavior would be if `body` was styled. It isn't _supposed_ to leak out, but in that case, it would have no effect.

The generated sample template uses an `<h1>` element. Perhaps that could be styled.

I don't feel strongly about it either, if you think making use of `:host` without explanation would be ok.

@kwalrath - thoughts?

kwalrath:
idk. Maybe @ferhatb or @matanlurey can recommend something?

chalin:
Actually, I'd say let's merge as is ... and if we get too many questions about `:host`, we can adjust either the CSS or add some prose somewhere.

kwalrath:
SGTM. Merging...

filiph:
Thanks Kathy!

PR #334, opened 2016-08-19
Add a component CSS file to the Angular template

filiph:
As proposed in issue https://github.com/google/stagehand/issues/332


kwalrath:
LGTM, though I should test this. We should also update CHANGELOG.md.

kwalrath:
On second thought... This has implications for the angular.io samples. @chalin might have an opinion.

filiph:
I just want to add that this change should make the tutorials easier and less error-prone. 

The step of creating a .css file and then adding its path to the component's metadata always felt like a chore to me.

chalin:
I agree about the inclusion of the style file, but see https://github.com/google/stagehand/issues/332#issuecomment-242222749 for a suggested alternative to `:host`.

kwalrath:
@filiph what do you have to say about `:host`?

kwalrath:
Merging, per discussion at #332.

PR #350, opened 2016-11-01
Add platform_* back to the web-angular pubspec.

### LINKED ISSUE #349: Angular2: add core directives and common pipes to angular2 pubspec options
Consider adding the following entries to the angular2 pubspec.yaml template.

This will make it easier for people to get started immediately.

```
- angular2:
    platform_directives: 'package:angular2/src/common/directives.dart#CORE_DIRECTIVES'
    platform_pipes: 'package:angular2/common.dart#COMMON_PIPES'
```

kwalrath:
We got feedback that removing the platform_* entries from the pubspec was a problem, so we're adding them back. Note that we're back to the original code, which is slightly different from what #349 suggested.

fixes #349


kwalrath:
@matanlurey perhaps you can review, since @kevmoo is swamped? (not that you're not...)

PR #353, opened 2016-12-01
Update web templates

kwalrath:
* add web-angular-quickstart
* remove web-polymer
* update web-angular

cc @chalin, @kevmoo


lib/generators/web_angular.dart
```diff
@@ -11,8 +11,8 @@ import 'web_angular_data.dart';
  */
 class WebAngularGenerator extends DefaultGenerator {
   WebAngularGenerator()
-      : super('web-angular', 'Angular 2 Web Application',
-            'A web app built using Angular 2.',
+      : super('web-angular', 'AngularDart Web Application',
```
chalin: We're not using "Dart" as a qualifier for "Angular" anywhere else. Is it really necessary here?

chalin:
@kwalrath - CI tests are failing because the new QS template is missing dot files (you can just copy the `web-angular` ones).

While we are at it, could we change the `web-angular/lib/app_component.html` to:
- At least drop the "2" from "Angular 2"?
- I think that it makes sense to have this mimic the QS template and use
`<h1>Hello {{name}}</h1>`, with at least the use of interpolation vs. _only_ static HTML.

lib/generators/web_angular.dart
kwalrath: I wanted to catch people's eye, but not I'm not sure it's necessary.

kwalrath:
I'm inclined not to change web-angular yet, since I hope we'll have only one web-angular template.

I've changed AngularDart to Angular in the quickstart description.

README.md
```diff
@@ -17,8 +17,8 @@ Kit and Yeoman.
 * `console-simple` - A simple command-line application.
 * `package-simple` - A starting point for Dart libraries or applications.
 * `server-shelf` - A web server built using the shelf package.
-* `web-angular` - A web app built using Angular 2.
-* `web-polymer` - A web app built using polymer.dart.
+* `web-angular` - A web app built using the latest version of Angular.
+* `web-angular-quickstart` - A minimalist example app used in docs.
```
devoncarew: We've used `-full` and `-simple` when we've wanted to have smaller and larger example apps. How attached are you to the `web-angular-quickstart` name? Do you think we should have consistency here - rename everything to `-simple` or `-quickstart` - or are ok with slightly different naming?

lib/generators/web_angular_quickstart_data.dart
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2014, Google Inc. Please see the AUTHORS file for details.
```
devoncarew: We should be generating this year's date. Probably a change to the generator code - I'll take a look.

templates/web-angular-quickstart/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: __projectName__
+description: QuickStart
```
devoncarew: Perhaps a longer description, similar to the one used above?

devoncarew:
lgtm w/ some changes noted. I can't speak to the angular code, but the stagehand mechanics look good :)

README.md
kwalrath: This example is a copy of the QuickStart example at https://angular.io/docs/dart/latest/quickstart.html.

I'm creating this template so it'll be easy and Dart-y to download the code (no need for GitHub, e.g.) so I can improve the (not-yet-published) Dart version of this page: https://angular.io/docs/ts/latest/guide/setup.html. 

Eventually, I hope that web-angular-quickstart will go away, and we'll just use web-angular.

So... I actually wanted an awkward name no one would accidentally type. :)

lib/generators/web_angular_quickstart.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2015, Google Inc. Please see the AUTHORS file for details.
```
kwalrath: Good catch. Fixed in my version.

templates/web-angular-quickstart/pubspec.yaml
kwalrath: Again, I'm just going with what's in the current example code. I wasn't even sure about using `__projectName__`, but decided to use it for now.

lib/generators/web_angular_quickstart_data.dart
kwalrath: OK, thanks.

PR #354, opened 2016-12-01
have the generated code use the current calendar year

devoncarew:
@kwalrath


kwalrath:
Thanks for the fix, @devoncarew.

PR #356, opened 2016-12-15
Add <meta> tags

filiph:
Add the two most common meta tags:

* charset for server compatibility (<1% issue)
* viewport for better scaling on mobile


filiph:
@kwalrath Can you take a look please? This should be a quick review. My next work on the AngularDart example is building on this.

templates/web-angular/web/index.html
```diff
@@ -1,8 +1,12 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta charset="utf-8">
+
```
kwalrath: Why so many blank lines? I usually leave them out, to save space when quoting code. Perhaps we could use this (mimicking https://github.com/google/stagehand/blob/master/templates/web-angular-quickstart/web/index.html):

```
...
  <head>
    <title>__projectName__</title>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <script defer src="main.dart" type="application/dart"></script>
    <script defer src="packages/browser/dart.js"></script>
  </head>
...
```

filiph:
Good point. PTAL.

PR #357, opened 2016-12-15
Add components to the AngularDart template

filiph:
The idea is to get people started with something simple yet functional and nice-looking.

stagehand-angulardart-app

This adds dependency on the official [angular2_components package](https://pub.dartlang.org/packages/angular2_components), and implements a simple app with `<material-input>`, `<material-button>` and `<material-dialog>`.

I'd love to get review from @kwalrath and @nshahan.


nshahan:
I'm clueless about these templates. How would I tell Intellij to load this as a template when I make a new project?

devoncarew:
@nshahan, it's already built into IntelliJ, via the new project wizard. It drive the stagehand cli behind the scenes.

nshahan:
@devoncarew But how would I get the code from this pull request into Intellij to try it out before it is merged?

filiph:
That's a good question. I don't think there's an easy way (you'd have to checkout the PR, then open the /template/web-angular project in IntelliJ or run `dart bin/stagehand.dart web-angular`).

If that's too much work (and it probably is), I can work on sharing this via GOB or something. But that'll take some time.

In the meantime, just from looking at the gif and the code — did I go overboard with complexity? I have second thoughts. Maybe using `<material-dialog>` is too much.

templates/web-angular/pubspec.yaml
```diff
@@ -5,6 +5,7 @@ environment:
   sdk: '>=1.19.0 <2.0.0'
 dependencies:
   angular2: ^2.2.0
+  angular2_components: ^0.2.2
```
nshahan: I don't think these version constraints are safe for angular2 and angular2_components. I would lock them down to the middle number so that the template works in the window between the release of a breaking change in angular and the release of the components to support it.

nshahan:
Ok thanks. I got it up and running. I don't think it is too complicated but what do you think is the main use case for the template? 

- If we want users to start work on their own applications with it, they would probably immediacy delete the contents of the html file.
- If we want users to just load it up and start playing around to learn more about angular and the components then it seems like a good starting place.
- If we want to use it as the starting point for a tutorial, maybe we should base it off the first steps?

templates/web-angular/web/index.html
```diff
@@ -5,8 +5,17 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <script defer src="main.dart" type="application/dart"></script>
-    <script defer src="packages/browser/dart.js"></script>
+    <script async src="main.dart" type="application/dart"></script>
```
kwalrath: Why async instead of defer?

templates/web-angular/lib/app_component.html
```diff
@@ -1 +1,36 @@
...
+<modal [visible]="helloVisible">
+  <material-dialog class="basic-dialog">
+
+    <h3 header>
+        Hello, {{ name == "" ? "mysterious stranger" : name }}!
+    </h3>
+
+    <p>
+      Continue your journey on
+      <a href="https://webdev.dartlang.org/angular">webdev.dartlang.org/angular</a>.
```
kwalrath: yay!

templates/web-angular/lib/app_component.dart
```diff
@@ -2,9 +2,18 @@
...
+import 'package:angular2_components/angular2_components.dart';
 
 @Component(
     selector: 'my-app',
     styleUrls: const ['app_component.css'],
-    templateUrl: 'app_component.html')
-class AppComponent {}
+    templateUrl: 'app_component.html',
+    directives: const [materialDirectives],
+    providers: const [materialProviders])
```
kwalrath: Should we use a terminal `,`? (We were doing that on purpose somewhere, but I don't remember the circumstances.)

templates/web-angular/lib/app_component.html
```diff
@@ -1 +1,36 @@
-<h1>My First Angular 2 App</h1>
+<h1>My First AngularDart App</h1>
+
+<material-input label="Your Name"
```
kwalrath: Name -> name

kwalrath:
> If we want users to start work on their own applications with it, they would probably immediacy delete the contents of the html file.

That's a possibility, and that's OK by me.

> If we want users to just load it up and start playing around to learn more about angular and the components then it seems like a good starting place.

That's the case we want to make easier.

> If we want to use it as the starting point for a tutorial, maybe we should base it off the first steps?

I'd like to eventually make this the starting point, but it's going to take some time. Until then, we'll keep the web-angular-quickstart template, as used in https://webdev.dartlang.org/angular/guide/setup.

Sfshaza:
@nshahan, async vs defer:
https://github.com/dart-lang/www.dartlang.org/issues/1526

templates/web-angular/pubspec.yaml
filiph: Thanks, I wasn't sure at all. So:

```
angular2: '>=2.2.0 <2.3.0'
angular2_components: '>=0.2.2 <0.3.0'
```

Correct?

templates/web-angular/lib/app_component.html
filiph: Ack

filiph:
> > If we want users to just load it up and start playing around to learn more about angular and the components then it seems like a good starting place.

> That's the case we want to make easier.

I have the same feeling, yes. On the other hand, I have second thought about the whole dialog thing. It makes `app_component.html` 2-3 times longer than it could be. And it may raise a lot of questions to the beginning user.

I'll experiment tomorrow. Thanks for all the feedback! Keep it coming.

templates/web-angular/lib/app_component.dart
nshahan: I liked the terminal `,`when I first saw it because it means in the future adding additional lines will have a very clean diff since you don't need to add the trailing comma.

It's a small thing. Don't know if there is a better argument for it out there.

templates/web-angular/pubspec.yaml
nshahan: I think those constraints look good. For now we should keep the versioning pretty narrow.

When Angular releases their next breaking changes and the components get updated to be compatible, they will bump to 0.3.0. At that time they will be incompatible with Angular 2.2.x.

filiph:
I reworked the template. @nshahan, could you please take another look? @ferhatb, would you mind looking at it, too?

The functionality is the same, but I wanted to show more best practices:

* Put anything non-trivial into its own component.
* Components in separate directories.
* Use #references and `@ViewChild` to simplify code.
* Call components’ functions.
* Put CSS as close to its use as possible.
* Use trailing commas to simplify moving code around.

Motivation: I see people (including myself, tbh) struggling with getting up to speed with new AngularDart projects. Like: "how was that annotation called, something with Ref?" or "where do I put the separate component?" or "how do I most easily get input from this field and put it over there?".

So I put that stuff right there in the template.

I'm okay with people deleting half of the template when they start a new project and don't need it. True, it could potentially get annoying if people who know AngularDart inside-out generate a lot of new projects and have to spend time deleting stuff every time. But I don't think that's something we should optimize for right now.

WDYT?

templates/web-angular/pubspec.yaml
```diff
@@ -4,8 +4,8 @@ version: 0.0.1
 environment:
   sdk: '>=1.19.0 <2.0.0'
 dependencies:
-  angular2: ^2.2.0
-  angular2_components: ^0.2.2
+  angular2: '>=2.2.0 <2.3.0'
```
nshahan: After our conversation offline, we could let angular2_components set the version constraints for the angular2 package.

templates/web-angular/lib/hello_dialog/hello_dialog.css
```diff
@@ -0,0 +1,3 @@
+a {
+  text-decoration: none;
```
nshahan: Why no decoration? Maybe add custom colors for the webdev.dart.org link.

nshahan:
I like the idea of showing a custom component in the template as a simple example.

templates/web-angular/pubspec.yaml
filiph: In the end, I realized that there's still a small chance that angular 2.3.0 _won't_ break angular2_components but _will_ break the sample app. To be on the safe side, I'm leaving this as is, with y-constraints on both packages.

templates/web-angular/lib/hello_dialog/hello_dialog.css
filiph: Ok, `:host` doesn't work here as I expected 😞 . The snippet above does nothing because it constraints `<hello-dialog>`'s width, and `<material-dialog>` doesn't care about the parent's width. So I'm back to `text-decoration`. If you feel strongly about that one, please let me know.

templates/web-angular/lib/app_component.dart
filiph: Yes, in the end I think the terminal `,` is better, and that's what the sample code uses now. Thanks for the suggestion, Kathy!

filiph:
@nshahan Could you give this another pass, please? I'd like to land this by the end of this week, if possible.

templates/web-angular/lib/app_component.dart
```diff
@@ -2,9 +2,17 @@
...
+import 'package:__projectName__/hello_dialog/hello_dialog.dart';
 
 @Component(
     selector: 'my-app',
     styleUrls: const ['app_component.css'],
-    templateUrl: 'app_component.html')
-class AppComponent {}
+    templateUrl: 'app_component.html',
+    directives: const [materialDirectives, HelloDialog],
+    providers: const [materialProviders],
```
a14n: With a trailing coma _dartfmt_ uses 2 spaces indentation

templates/web-angular/lib/app_component.dart
```diff
@@ -2,9 +2,17 @@
 // is governed by a BSD-style license that can be found in the LICENSE file.
 
 import 'package:angular2/core.dart';
+import 'package:angular2_components/angular2_components.dart';
+
```
a14n: _Optimize imports_ in IntelliJ removes this empty line.

templates/web-angular/lib/hello_dialog/hello_dialog.html
```diff
@@ -0,0 +1,22 @@
...
+    <p>
+      Continue your journey on
+      <a href="https://webdev.dartlang.org/angular">webdev.dartlang.org/angular</a>.
+    </p>
+
+    <div footer>
+      <material-button autoFocus
+                       clear-size
+                       (trigger)="wrappingModal.close()"
+                       class="white">
```
a14n: Unlike `blue` in `lib/app_component.html` I can't find `white` css class. Where is it from ?

templates/web-angular/lib/app_component.dart
filiph: Good catch, ack!

templates/web-angular/lib/app_component.dart
filiph: True, but the de-facto Dart style is to have a blank line between third party and first party imports (and then another one between package: and relative ones). See [here](https://www.dartlang.org/guides/language/effective-dart/style#ordering). It doesn't really matter for most code, but sample code especially should go the extra mile to be super readable.

Unless you feel strongly about this, I'd keep this the way it is.

templates/web-angular/lib/hello_dialog/hello_dialog.html
filiph: Damn. Great catch! Ack.

templates/web-angular/lib/app_component.dart
a14n: Related issue : dart-lang/sdk#26020

templates/web-angular/lib/hello_dialog/hello_dialog.dart
```diff
@@ -0,0 +1,27 @@
...
+
+@Component(
+  selector: 'hello-dialog',
+  styleUrls: const ['hello_dialog.css'],
+  templateUrl: 'hello_dialog.html',
+  directives: const [materialDirectives],
+  providers: const [materialProviders],
+)
+class HelloDialog {
+  /// Modal component that hoists the inner MaterialDialog in a centered overlay.
```
kwalrath: Not sure what "hoists" means here. Did you mean "hosts"? Perhaps choose a clearer word.

templates/web-angular/lib/hello_dialog/hello_dialog.dart
```diff
@@ -0,0 +1,27 @@
...
+  @ViewChild('wrappingModal')
+  ModalComponent wrappingModal;
+
+  /// Name of user.
+  @Input()
+  String name = "";
```
kwalrath: "" -> ''

templates/web-angular/lib/hello_dialog/hello_dialog.html
```diff
@@ -0,0 +1,21 @@
+<modal #wrappingModal>
+  <material-dialog class="basic-dialog">
+
+    <h3 header>
+        Hello, {{ name == "" ? "mysterious stranger" : name }}!
```
kwalrath: I'm not sure whether to use our site's HTML convention here (double quotes) or our Dart convention (single quotes).

templates/web-angular/web/index.html
kwalrath: Did you get an answer? We went and changed all our asyncs to defers a while ago. I think @sigmundch was involved in that conversation.

kwalrath:
LGTM. A few nits, but once you've resolved them go ahead and merge the PR, and I'll make a stagehand release.

templates/web-angular/lib/hello_dialog/hello_dialog.dart
filiph: What's that stand for? Should be on one line?

templates/web-angular/lib/hello_dialog/hello_dialog.html
filiph: I'll rewrite to `'`.

templates/web-angular/web/index.html
filiph: No reply yet. I'm switching back to defer to be safe. We can update this later.

kwalrath:
About to create a new release of stagehand, so people will get these changes when they restart their IDEs.

kwalrath:
Release is out: 1.0.20

nshahan:
I just tested in IntelliJ. Worked great.

PR #361, opened 2017-02-06
upgrade to usage 3.0.0

devoncarew:
- upgrade to usage 3.0.0

@kwalrath


kwalrath:
LGTM. Do we need a technical review from anyone?

devoncarew:
@kevmoo could take a look, but this is pretty minor stuff.

devoncarew:
I'm assuming this isn't really important enough for its own release - that it would be picked up with the next general publish.

PR #362, opened 2017-02-13
rename the .analysis_options files to analysis_options.yaml

devoncarew:
- rename the .analysis_options files to analysis_options.yaml
- part of https://github.com/dart-lang/sdk/issues/28385

@danrubel @kevmoo


kwalrath:
Are we going ahead with the name change, despite the numbers Bob quoted in dart-lang/sdk#28385? (I suppose we can make this change here, even if we don't make it elsewhere.)

danrubel:
My plan is to push ahead with the name change, although it will be a slow process.

Issue #360, opened 2016-12-30
Add <base href="/"> to index.html in an Angular project


kulshekhar:
Currently this needs to be added manually whenever routing is needed and it would be quite convenient if this were present in the template generated by Stagehand, at least in the `web-angular` template if not also in the `web-angular-quickstart` template

kwalrath:
Thanks for filing this issue, @kulshekhar.

@chalin @filiph @matanlurey: thoughts?

matanlurey:
Sgtm

chalin:
This has been addressed by #365 and can be closed.

PR #363, opened 2017-02-15
Added <base href="/"> to index.html in angular templates (#360)

switchspan:
Hi! Saw this in issues and figured it was a quick fix. I was 'bitten' by the absence of this tag on the tour of heroes tutorial's routing section: https://webdev.dartlang.org/angular/tutorial/toh-pt5 (don't worry -- I recovered without a scar)


kwalrath:
@chalin what do you say?

chalin:
@switchspan, thanks for the update.

@kwalrath - I think that I'd like to see the following right after the `<base href>` element:
```
    <!-- For testing in WebStorm use: -->
    <!-- base href="/dart/web/" -->
```
We had this comment we have in the toh-5 `index.html`. WDYT?  If Kathy agrees, can you make that update @switchspan?

switchspan:
@chalin No prob -- I can update it. @kwalrath what say you?

kwalrath:
Sounds good to me. Thanks, @switchspan & @chalin!

switchspan:
@kwalrath @chalin No problem. I added the comments and re-ran a build from grind (almost forgot this :-o ). It should be good to go now.

kwalrath:
My only question... Should we really be adding 3 lines to web-angular-quickstart? Since it doesn't do any routing and this is supposed to be the minimum possible app, maybe this should be handled in toh-5 directly?

switchspan:
@kwalrath Good point. I can definitely see an argument for adding it to toh-5 so that others following it are aware that when using routing the base tag must be present, or APP_BASE_HREF needs to be imported & provided.

Should I remove it from quickstart and we punt that part over to whatever repo has the toh-5 content?

kwalrath:
Yes, let's do that. Thanks VERY much for your responsiveness!

While you're in there, could you update the CHANGELOG and put the
#unreleased item under 1.0.21? (You did the right thing, they technically
didn't. :) I plan to release 1.0.21 as soon as your change is in.

switchspan:
@kwalrath No problem. I can do that. Also -- can you point me to the right repo to submit an update for the toh-5 docs?

kwalrath:
toh-5 docs & examples are all in https://github.com/dart-lang/site-webdev.

The webdev build system is a bit gnarly, so if it gets too onerous, just tell me what to change and I'll make it happen. Thank you!

switchspan:
@kwalrath Thanks -- Will do! The lastest changes are now commited to the pr branch.

chalin:
Hey all. I'm in the middle of something. Could we just put a hold on this for a little while. I'll need to discuss some broader issues with @kwalrath offline. Thanks.

switchspan:
@chalin Sure. I will hold up on any submissions of code/issues to site-webdev til I hear from you.

kwalrath:
Patrice and I have talked, and we're going to keep base href in web-angular-quickstart. But I'm going to make that change myself, since I'm going to change the name of that template anyway.

I believe Patrice is midway through his router changes, so hold off on those...

Thanks very much for your PR!

kwalrath:
Actually, since base href makes the initial WebStorm experience bad, I'm rethinking how we'll handle it.

switchspan:
@kwalrath I just ran it under WebStorm and I can see what you are talking about. :-(
One path might be removing it from the quickstart and adding routing to the regular angular template -- but that might be overkill if we don't need routing right away. I will take a look and see what yeoman is creating template-wise for a full angular 2 project.

kwalrath:
We're thinking about adding something like this just after the `<head>` tag:

```
    <!-- Set the base href dynamically so that routing works with most servers.
         Don't do this in production!
         Details: webdev.dartlang.org/angular/guide/router -->
    <script>document.write('<base href="' + document.location + '"/>');</script>
```

kwalrath:
See #365 for related changes.

PR #364, opened 2017-02-15
Update LICENSE Copyright Date to 2017

Phineas:


Phineas:
I signed it!

kwalrath:
I had to look into this, since copyright policies vary. It turns out that we don't usually update the copyright year, and instead just use the current year when we create a new file.

Thanks very much for the PR, though!

PR #365, opened 2017-02-16
web-angular* template changes, prep for 1.0.21

kwalrath:
THIS CHANGE AFFECTS DOCS. `web-angular-quickstart` is now `web-angular-simple`, and the description of `web-angular` has changed slightly.

I'll need to update WebStorm screenshots and anything that contains the string `web-angular-quickstart`.

I plan to release 1.0.21 as soon as those doc changes are ready to go.


chalin:
LGTM

devoncarew:
lgtm

devoncarew:
You'll need to update stagehand.io after this PR lands -

chalin:
Actually, after further testing, it seems that this doesn't play will with servers that do rewrites (like rewriting all URLs to index.html).

On the one hand, we don't officially have a recommended server yet that does this kind of rewriting (I'm using a custom version of `pub serve`).  So maybe this point is moot?

Anyhow, I'm looking for alternatives.

chalin:
Here is the script I propose that we use. It works under WebStorm and when using pub serve, the two approaches we recommend in the docs.
```js
    <script>
      var _ngBaseHref = document.location.pathname.startsWith('/dart/web/') ? '/dart/web/' : '/';
      document.write('<base href="' + _ngBaseHref + '" />');
    </script>
```
Btw, it seems that this would address #360.

kwalrath:
That code doesn't work for me. Perhaps my path in WebStorm is different from your path? I tried restarting... no luck.

image

chalin:
I think that we might need to look at your WS config. I don't think that the default path prefix is `/was/web/`.

kwalrath:
I think it's _your_ WS config that's unusual (perhaps because it uses your version of pub serve?).

`~/tmp/was` was a directory I created by hand to test the web-angular-simple template. I created the files using stagehand, and then opened the directory in WebStorm.

I just tried creating an app in WebStorm, and then copying in your base href-setting code. I get the same result.

image

chalin:
Ah, thanks for the insight!

This should work for you now:
```html
    <script>
      // WARNING: DO NOT set the <base href> like this in production!
      // Details: https://webdev.dartlang.org/angular/guide/router
      (function () {
        // App being served out of web folder (like WebStorm does)?
        var match = document.location.pathname.match(/^\/[_\w]+\/web\//);
        var href = match ? match[0] : '/';
        document.write('<base href="' + href + '" />');
      }());
    </script>
```

kwalrath:
Yes, that works! Thanks!

kwalrath:
I've updated the templates with Patrice's new code.

filiph:
Hi, this breaks the IntelliJ user experience completely when you use `-` (or any other such char) in the project name.

See [WebStorm issue in YouTrack](https://youtrack.jetbrains.com/issue/WEB-25766).

The quick fix is (I believe) to change

    var match = document.location.pathname.match(/^\/[_\w]+\/web\//);

to

    var match = document.location.pathname.match(/^\/[_\-\w]+\/web\//);

But I haven't tested this, and it all seems very brittle and inelegant.

Is this the only way for us to support routing? Wouldn't it be easier on the developer to just say something like:

    
    
    

Of course, that would force us to educate people about this. But the starting experience, when you don't even need routing yet, is now unnecessarily broken because of this.

chalin:
A **project** (or more generally a package) [**name** must be a legal Dart identifier](https://www.dartlang.org/tools/pub/pubspec#name), and hence `-` is prohibited.

Usually, it makes the most sense to have a project _folder_ name that matches the project name in the `pubspec.yaml`. This is why the Regex is what it is.

We originally had just a comment but we wanted a solution that worked out-of-the-box for our two "officially supported" scenarios: WebStorm/IJ and `pub serve` for command line.

filiph:
> A project (or more generally a package) name must be a legal Dart identifier, and hence a hyphen (-) is prohibited.

I know, but that doesn't apply for WebStorm projects. I could easily create a Dart project with a dash in it, and WebStorm nor Dart analyzer complained about it. The result is broken experience for anyone who happens to use `-` in their project.

We could 'solve' this by adding a note in the documentation, but the getting started experience already has too much of those, I'm afraid. Are there other ways to address it?

chalin:
I agree with your suggested quick fix from https://github.com/google/stagehand/pull/365#issuecomment-285537758 -- i.e., adding `-` to the list of permitted characters in the project name pattern match.

filiph:
Before I start the PR: is there any other type of symbol that I should include in that RegExp? Something that people on Mac/Linux/Windows often put into their project folder names and that isn't covered by `[_\-\w]`?

filiph:
Btw, it seems that `_` is already covered by `\w`:

> A word character is a character from a-z, A-Z, 0-9, including the _ (underscore) character.

kwalrath:
> Before I start the PR: is there any other type of symbol that I should include in that RegExp? Something that people on Mac/Linux/Windows often put into their project folder names and that isn't covered by `[_\-\w]`?

I can't think of anything.

chalin:
Yes, `_` is already in the "identifier characters" group `\w`. I'd leave it out.

PR #367, opened 2017-02-23
templates/web-angular-simple: add copyright notice + formatting

chalin:
cc @kwalrath


templates/web-angular-simple/lib/app_component.dart
```diff
@@ -1,6 +1,12 @@
+// Copyright (c) __year__, __author__. All rights reserved. Use of this source code
+// is governed by a BSD-style license that can be found in the LICENSE file.
+
```
kwalrath: Hmmm. It looks like we need to add a LICENSE file to this and web-angular.

kwalrath: We should also add CHANGELOG.md and README.md to web-angular, but perhaps not here.

chalin: Done and done.

PR #369, opened 2017-02-23
templates/web-angular: use global styles.css, hello dialog tweak

chalin:
- Use global `styles.css`, vs inlined styles
- `hello_dialog` tweak to support next revision of [webdev/ng "Get Started"](https://webdev.dartlang.org/guides/get-started).

cc @filiph @kwalrath


kwalrath:
LGTM. @filiph let us know if you disagree with any of these changes. Related PR: https://github.com/dart-lang/site-webdev/pull/406.

chalin:
> we should figure out a way to automatically inline the global css on build.
Good idea to do that as part of the build.

kwalrath:
@filiph perhaps you could create an issue requesting this auto inlining. Would it belong in https://github.com/dart-lang/bazel?

Issue #372, opened 2017-03-10
Update pubspecs to use ^ syntax


kevmoo:
Make the min SDK 1.8.3 – this is the first release where the ^ syntax is valid

Update the dependencies to use this syntax (just not for SDK – for backwards compat)

PR #373, opened 2017-03-10
Big cleanup to .gitignore and pubspec.yaml for all templates

kevmoo:
- Standardized the layout of `pubspec.yaml` for all templates.
  - Consistent ordering of all top-level keys
  - Use `^` notation for versions.
  - Define the minimum SDK to be `1.20.1`.
- Simplified `.gitginore` for all templates.
  - No longer need to ignore `packages` directory. They are not not generated
    with SDK 1.20.1+.
  - No longer ignoring compiled outputs by extension. These will all be in the
    `build` directory, which is already ignored.
  - No longer ignoring JetBrains IDE files. These should be ignored in a users
    global `.gitignore`.

Also added tests to validate many of these things

Fixes https://github.com/google/stagehand/issues/372


templates/console-full/pubspec.yaml
```diff
@@ -1,14 +1,14 @@
...
 #homepage: https://www.example.com
+#author: __author__ <email@example.com>
 
 environment:
-  sdk: '>=1.0.0 <2.0.0'
+  sdk: '>=1.20.1 <2.0.0'
 
 #dependencies:
-#  foo_bar: '>=1.0.0 <2.0.0'
+#  path: '^1.4.1'
```
kwalrath: Do we really want the `'`s? It's so clean without them.

templates/web-angular-simple/pubspec.yaml
```diff
@@ -1,13 +1,19 @@
...
 version: 0.0.1
+#homepage: https://www.example.com
+#author: __author__ <email@example.com>
+
 environment:
-  sdk: '>=1.19.0 <2.0.0'
+  sdk: '>=1.20.1 <2.0.0'
+
 dependencies:
   angular2: ^2.2.0
```
kwalrath: If we want to use `'`s in the version strings, then this needs `'`s.

templates/web-angular-simple/pubspec.yaml
```diff
@@ -1,13 +1,19 @@
...
+
 dev_dependencies:
   browser: ^0.10.0
   dart_to_js_script_rewriter: ^1.0.1
```
kwalrath: ditto for these version strings.

templates/console-full/pubspec.yaml
devoncarew: +1 to using the `^` operator w/o quotes

templates/console-simple/pubspec.yaml
```diff
@@ -1,5 +1,14 @@
 name: __projectName__
-version: 0.0.1
 description: A simple console application.
+version: 0.0.1
+#homepage: https://www.example.com
+#author: __author__ <email@example.com>
+
+environment:
+  sdk: '>=1.20.1 <2.0.0'
```
devoncarew: The min version number seems somewhat arbitrary?

devoncarew: nevermind, you mention it's the version w/o `packages` dirs

devoncarew:
lgtm

templates/console-full/pubspec.yaml
chalin: +1 for `^` w/o quotes too

templates/web-angular-simple/pubspec.yaml
kevmoo: yup

templates/web-angular-simple/pubspec.yaml
kevmoo: We have to use 'em here – because there are spaces.

templates/console-full/pubspec.yaml
kevmoo: done

Issue #374, opened 2017-03-10
change web-angular* templates to allow - in the project name


kwalrath:
See https://github.com/google/stagehand/pull/365#issuecomment-285537758.

We'll need corresponding changes to the site-webdev Angular examples.

filiph:
When's the next release of stagehand? People are bumping into this issue (https://github.com/google/stagehand/issues/379).

kwalrath:
I'll release it right now.

kwalrath:
And it's out: 1.0.24

filiph:
Thank you Kathy!

PR #376, opened 2017-03-11
Support WebStorm project names with hyphens

filiph:
Fixes https://github.com/google/stagehand/issues/374 (not the site-webdev documentation part).

Note that `_` was removed from the regexp because `_` is already included by `\w`.


filiph:
Thanks for the feedback! I added to web-simple and simplified the regexp per suggestion.

PR #382, opened 2017-03-20
Updated analysis_options: new defaults such as strong-mode

### LINKED ISSUE #368: Angular template: in analysis options, enable strong-mode
Cf. https://github.com/google/stagehand/blob/master/templates/web-angular/analysis_options.yaml#L6

@matanlurey @filiph @devoncarew @kevmoo - is there any reason _not_ to do this?

cc @kwalrath

chalin:
Fixes #368

I ran the analyzer over all samples (using the new options) and now errors or warnings are reported.

cc @kwalrath @filiph @kevmoo


kevmoo:
Are we sure we like this set of things? Why was this set picked?

chalin:
This lint rules are from @filiph's "most wanted". The list was briefly discussed and then agreed to in #368.

chalin:
Updates done, and Travis is green.

test/common_test.dart
```diff
@@ -27,10 +27,10 @@ void main() {
...
 
       test('vars must be alpha + numeric', () {
-        expect(() => substituteVars('str', {'with space': 'noop'}), throws);
-        expect(() => substituteVars('str', {'with!symbols': 'noop'}), throws);
-        expect(() => substituteVars('str', {'with1numbers': 'noop'}), throws);
-        expect(() => substituteVars('str', {'with_under': 'noop'}), throws);
+        expect(() => substituteVars('str', {'with space': 'noop'}), throwsA(anything));
+        expect(() => substituteVars('str', {'with!symbols': 'noop'}), throwsA(anything));
+        expect(() => substituteVars('str', {'with1numbers': 'noop'}), throwsA(anything));
+        expect(() => substituteVars('str', {'with_under': 'noop'}), throwsA(anything));
```
kevmoo: What is actually thrown? Let's be explicit here. There's a reason `throws` was deprecated. 😄

test/validate_templates.dart
```diff
@@ -27,14 +27,15 @@ final _pubspecOrder = const [
...
+final String _expectedGitIgnore =
+    new File(path.join(path.current, '.gitignore')).readAsStringSync();
+final String _expectedAnalysisOptions =
+    new File(path.join(path.current, 'analysis_options.yaml'))
+        .readAsStringSync();
 
 void main() {
   Directory dir;
 
-  _expectedGitIgnore =
```
kevmoo: I had this here for a good (annoying) reason. If this fails on access (like the first access of the field) then the first test will fail w/ a File exception...then every other test will fail w/ a NPE.

Setting this one is main ensures the whole test blows up, which is likely easier to debug

test/common_test.dart
chalin: Fixed (they throw `ArgumentError`).

test/validate_templates.dart
chalin: Ok. I think that I've addressed your concern.
If not, then let me know and I'll roll back the changes.

test/common_test.dart
```diff
@@ -26,11 +26,12 @@ void main() {
             'foo __baz__ baz');
       });
 
+      const _argError = const isInstanceOf<ArgumentError>();
```
kevmoo: https://www.dartdocs.org/documentation/test/0.12.20%2B3/test/throwsArgumentError-constant.html 

😄

test/validate_templates.dart
```diff
@@ -27,11 +27,9 @@ final _pubspecOrder = const [
 final List<RegExp> _pubspecOrderRegexps =
     _pubspecOrder.map((s) => new RegExp('^#?$s:', multiLine: true)).toList();
 
-final String _expectedGitIgnore =
-    new File(path.join(path.current, '.gitignore')).readAsStringSync();
+final String _expectedGitIgnore = _getMetaTemplateFile('.gitignore');
```
kevmoo: Try deleting the file and running the tests and see what happens.

😄

test/common_test.dart
chalin: Ah, good to know.

test/validate_templates.dart
chalin: I did. With the current setup:
- (a) The "meta-template file exists" test will fail if either of the meta-template files are missing (and all other tests will fail reporting the the corresponding template file doesn't match `''`).
- (b) If they are present, and the templates have mismatching file content, then we'll get specific failures for each template.

That seems ok to me, esp. given that (a) is unlikely to happen often. Let me know if this is agreeable with you. If not, how I can make it better.

kevmoo: Cool cool

test/common_test.dart
chalin: Fixed

kevmoo:
Thanks so much, @chalin !!

PR #397, opened 2017-05-05
[WIP] web-angular template: add tests, enhance demo

### LINKED ISSUE #396: web-angular: rework template and add basic component tests
Part of #390.

chalin:
Fixes #396


templates/web-angular/lib/app_component.dart
```diff
@@ -4,15 +4,35 @@
...
+    name = whoToGreet.isEmpty ? 'mysterious stranger' : whoToGreet;
+  }
+}
+
+Iterable<String> buildNames() sync* {
+  // Suggested function body
+  // (see https://webdev.dartlang.org/guides/get-started for details):
+  /*
+    var prefixes = ['Super', 'Mega', 'Ultra'];
+    var suffixes = ['man', 'woman', 'boy', 'girl'];
```
kwalrath: Good catch that the DartPad uses " when we usually use '!

templates/web-angular/lib/app_component.dart
```diff
@@ -4,15 +4,35 @@
...
+  String name = '';
+  List<String> heroes = buildNames().toList();
+
+  void greet(String whoToGreet) {
+    name = whoToGreet.isEmpty ? 'mysterious stranger' : whoToGreet;
+  }
+}
+
+Iterable<String> buildNames() sync* {
+  // Suggested function body
```
filiph: I'd just put 

```
Iterable<String> buildNames() sync* {
  // See https://webdev.dartlang.org/guides/get-started for suggested function body.
  yield "Ultraboy";
  yield "Megagirl";
}
```
I'm purposely using `yield` and not using `yield*` here to keep things simple yet interesting. But I don't feel strongly about it, so we might just use `return ["Ultragirl", "Megaboy"]`.

templates/web-angular/lib/app_component.html
```diff
@@ -1,17 +1,28 @@
...
-                autoFocus>
-</material-input>
+<div class="name-input">
+  <material-input #nameInput label="Your name" floatingLabel autoFocus>
+  </material-input>
 
-<material-button raised 
-                 (trigger)="myDialog.open()"
-                 class="blue">
+  <material-button raised (trigger)="greet(nameInput.inputText)" class="blue">
```
filiph: Can you put one attr per line? It's more readable.

templates/web-angular/web/index.html
```diff
@@ -1,15 +1,6 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <script>
```
filiph: Do you want to add a `<!-- comment -->` here with a link to the docs? I think it's fine without the comment (since people will read the docs, and the docs will hopefully always contain the info about the workaround) but just wanted to check.

templates/web-angular/pubspec.yaml
```diff
@@ -22,4 +24,8 @@ transformers:
     platform_pipes:
     - 'package:angular2/common.dart#COMMON_PIPES'
     entry_points: web/main.dart
+- angular2/transform/reflection_remover:
```
filiph: Ugh, I wish we didn't need things like this :/. Welp.

(No action necessary, just a rant.)

chalin:
Thanks for the feedback. I agree with dropping the modal dialog component, and the idea of simplifying the app.

I was aiming for a (somewhat) cohesive UI, but the greeting and "Hall of Fame" features don't quite relate, so I put them in separate tabs.

How about this: rather than **using the input+button** for a greeting, we could allow the user to **add heroes to the list** with it.  If you agree, then shall we keep the "_Hero Hall of Fame_" theme or just make it a list of "_My Favorite Heroes_"?

If you agree we'd have:
- AppComponent, and
- a hero list component

It would be nice if there was some other interactivity in the app. Here are some possibilities:
- We can still have a modal dialog w/o putting it in a component.
- We could make the hero list a re-orderable list.
- ... any other thoughts?

templates/web-angular/lib/app_component.html
chalin: Just like we use `dartfmt` for the Dart code, I use the Xcode HTML formatter for HTML files. It is much less of a hassle to maintain.

Formatting is such a thorny issue. It's great when the arbitrer is an automated formatter :).

In the light of this, let me know what you prefer.

templates/web-angular/lib/app_component.dart
chalin: If we opt to keep the heroes list, then I'll use the shorter code that you suggest.

templates/web-angular/web/index.html
chalin: Thanks for checking, but I'd vote to not have a comment.

templates/web-angular/lib/app_component.html
filiph: It was just a suggestion. I would normally also defer to an automated formatter but in this pedagogic case, I'm slightly leaning towards "nicer" formatting. The one-attr-per-line formatting is also something you can see in the AngularDart components docs, so it is somewhat a standard.

But again, I don't feel strongly about this.

templates/web-angular/web/index.html
filiph: 👍

filiph:
> If you agree we'd have:
> * AppComponent, and
> * a hero list component

Now that I see it like this, I wonder if we want to also have a Service, and put some basic functionality there. In that, we'd have a "full" AngularDart app. It could be very simple, like `class NameService { var names = ['John', 'Jane']; }` but it would still be in a separate file like it should. WDYT?

Otherwise, I agree, let's have an appcomponent and one other 'custom' component, and that's it.

> How about this: rather than using the input+button for a greeting, we could allow the user to add heroes to the list with it. If you agree, then shall we keep the "Hero Hall of Fame" theme or just make it a list of "My Favorite Heroes"?

Since we're thinking about changing behavior, I'd like to step back a little.

The original theme of this template was "Hello $name", which is more in line with what most people will be writing at first. I'm not sure if we want to force ourselves into the "Heroes" theme here. It's in the tutorial, but could seem out of place for a template.

So now I'm asking myself:

1. What does AngularDart with components shine in when it comes to building "weekend fun projects"? What kind of functionality is easiest to implement quickly?
2. What theme is self-explanatory enough that it seems natural or even "inevitable" for a template?

For 1), I'd say the "add <item> to list" functionality is a good fit. AngularDart makes it easy to do lists and update them. In contrast, the realtime update of text in DOM is nice for demos, but actually not that interesting.

For 2), I'd say hello world, but that's too easy and doesn't show _anything_ of AngularDart. Another would be a TODO list, which is almost as cliché as Hello World. Another is fibonacci, but that again doesn't show much of AngularDart (and also is computationally expensive, and you can bet people will try to compute 100000! in their browser, given the chance 😅 ). I can't think of anything else that close to Hello World.

So, I vote for the simplest possible TODO app that's also kind of complete.

* A service persists `TodoItem`s in memory.
* Main component has a list of TodoItems.
* The custom element is a `<todo-list-item>`.
* The main component has a `<material-input>`, an `<material-button>` with "Add", and a list of `<todo-list-item>`s.
* Material input allows adding new todo items.

The following functionality only if it doesn't make the code too complex. Sorted by priority:

* It would be great if we could rearrange them with the AngularDart component.
* Each todo item can be marked done, with a `<material-checkbox>`.
* Each todo item can be deleted, with a `<material-button>` and a "trash" glyph.
* Use localStorage only if that's 2-3 lines of additional code. Even then, I'm not sure we want this, because then every new AngularDart app will create / rewrite a localStorage record for localhost.

WDYT?

chalin:
Excellent reflections.

Note that the "heroes" theme tied in with the tutorial, but it was also there because the [Get Started](https://webdev.dartlang.org/guides/get-started) page starts off with a [DartPad code excerpt that generates hero names](https://dartpad.dartlang.org/d01f9074dc1001595159ef2b4eb68a0d). As you can recall, towards the end of the Get Started page, we have the reader make use of the DartPad code to enhance the `web-angular` project they created.

That being said, I like the idea of a to-do list, but then how do you suggest that we adapt the Get Started page? We could choose to not make use of the DartPad code in the app, or change the DartPad code excerpt. WDYT?

filiph:
I'd keep the hero dartpad sample in Get started as is. It's probably more
interesting that way, and ties in with the tutorial, as you say.

The stagehand template doesn't need to tie in with the tutorial, imho. I'm
fine if people delete the little code there is in the stagehand at the
start of the tutorial.

I'm sure there will be people who'd prefer completely blank template, but I
think the pedagogical factor of a 'full' app is too valuable.

chalin:
Ok. I'll revise the template along the lines you suggested.

chalin:
@filiph @kwalrath - the new PR is https://github.com/google/stagehand/pull/401. I decided to submit it before completing the tests. Please have a look and let me know what you think.

@filiph - after some experimentation, it seemed to work better to have the custom component be a `<todo-list>` rather than a list item.

chalin:
Superseded by #401.

PR #400, opened 2017-05-05
Delete web-angular-simple & console-simple templates

kwalrath:
Also tweak the descriptions of the remaining web-* templates, do a minor edit of the README,
and remove leftover polymer files.

Contributes to fixing #390.
Fixes #387.

Issue #390, opened 2017-04-15
curate the set of available templates


devoncarew:
Mentioned in https://github.com/google/stagehand/issues/389:

Another item we've encountered in UX reviews is there are are too many new project templates now (https://github.com/google/stagehand/tree/master/templates) - there's too much choice for the user and not enough clarity on which is the 'right' template. It was always a goal of stagehand to be proscriptive - to have a few high quality templates rather than supporting many different project types and configurations.

We'll likely need to drop a few templates and have clearer guidelines about when new ones are added.

devoncarew:
(we have 8 templates currently - I could see us getting down to 4-5 pretty easily)

kwalrath:
Agreed that there are too many to easily scan. It'd be nice if we could filter templates and/or make it easy to add new ones.

If you're developing a web app, you don't want to see the console app or shelf web server templates. Most people won't need the stagexl app template, but it'd be nice to plug it in if you do want it.

@kevmoo @filiph @chalin we should talk about the templates.

devoncarew:
> If you're developing a web app, you don't want to see the console app or shelf web server templates. 

Even if you're developing a web app, we have 4 web templates to choose from. That's likely too much choice for a new user, w/o the right context to decide between the available options.

filiph:
I see the problem mostly with Angular (2 ways) which I know is about to be addressed. So I vote for having:

* One "AngularDart Web App"
* One "Low-level HTML App"
* One "StageXL Game or App"

Note the changed wording.

kwalrath:
Related issue: #192

kwalrath:
I'm going to take a crack at this as soon as I've updated the docs to not depend on web-angular-simple (https://github.com/dart-lang/site-webdev/issues/591).

kwalrath:
Actually, @chalin is working on this first.

kwalrath:
Do we really need two console samples?

devoncarew:
> Do we really need two console samples?

+1 to dropping one; I haven't looked at those samples for a while, but I suspect generating a sample with a test would be useful guidance for people.

kwalrath:
@filiph, @mit-mit, @anders-sandholm, @kevmoo what do you think of dropping **console-simple** and changing **console-full** to plain old **console**?

filiph:
I do use `console-simple` for doing things like bug repros. But it's true that for "real" projects (even weekend projects) I'd rather go with `console-full` if I can only get one. So yes, LGTM.

[...]

PR #403, opened 2017-05-08
Delete web-angular-simple & console-simple templates

kwalrath:
Also tweak template descriptions.

This contributes to #390. This PR replaces #400, which was LGTM'd but also removed the StageXL template.


README.md
```diff
@@ -58,10 +56,10 @@ And to list all of the project templates:
 ## Goals
 
 * Opinionated and prescriptive; minimal to no options
-* Mobile-first
+* Mobile first
```
devoncarew: Hmm, I know this is existing, but how are we mobile first? Perhaps remove?

lib/generators/web_simple.dart
```diff
@@ -11,8 +11,8 @@ import 'web_simple_data.dart';
  */
 class WebSimpleGenerator extends DefaultGenerator {
   WebSimpleGenerator()
-      : super('web-simple', 'Simple Web Application',
-            'An absolute bare-bones web app.',
+      : super('web-simple', 'Low-level HTML App',
```
devoncarew: It's not really clear to me what the term `Low-level` means in the context of a web app; `Simple Web App` seems clearer.

kwalrath: @filiph?

README.md
kwalrath: How about "Mobile friendly"?

filiph: Stagehand is mobile friendly? If you mean that some of the templates are, then we should be clear on that — something like "Stagehand templates for the web are mobile first".

lib/generators/web_simple.dart
filiph: Hmm, to me, Low-level HTML App is actually clearer than "Simple Web App". How about "Web App without any framework"? We want to set the expectations that this is different from the AngularDart template.

site/index.html
```diff
@@ -91,12 +91,10 @@ <h1 id="usage">Usage</h1>
...
 <ul id="template-list">
   <li>console-full - <em>A larger command-line application sample.</em></li>
-  <li>console-simple - <em>A simple command-line application.</em></li>
   <li>package-simple - <em>A starting point for Dart libraries or applications.</em></li>
   <li>server-shelf - <em>A web server built using the shelf package.</em></li>
-  <li>web-angular - <em>A web app built using the latest stable version of Angular.</em></li>
-  <li>web-angular-simple - <em>A minimalist example app used in docs.</em></li>
-  <li>web-simple - <em>An absolute bare-bones web app.</em></li>
+  <li>web-angular - <em>A web app with material design components.</em></li>
+  <li>web-simple - <em>A bare-bones web app.</em></li>
```
filiph: A bare-bones web app without use of any framework? A framework-less web app?

kwalrath: How about:

```
Available generators:
  console-full   - A command-line application sample.
  package-simple - A starting point for Dart libraries or applications.
  server-shelf   - A web server built using the shelf package.
  web-angular    - A web app with material design components.
  web-simple     - A web app that uses no framework.
  web-stagexl    - A starting point for 2D animation and games.
```

kwalrath: And the (more important) IDE equivalents:

Console Application - a command-line application sample.
Dart Package - a starting point for Dart libraries or applications.
Shelf Web Server - a web server built using the shelf package.
AngularDart Web App - a web app with material design components.
Bare-bones Web App - a web app that uses no framework.
Simple StageXL Application - a starting point for 2D animation and games.

kwalrath: Hmmm... the last one could be:

StageXL Web App - ...

kwalrath: cc @Tomucha

chalin: One might argue that `web-simple` isn't really what we'd want to call a "web app". Anyhow ...

Rather than saying what it doesn't use, why not say what it does use? E.g. "A web app using only core Dart libraries"

kwalrath:
@devoncarew & @filiph take another look, if you dare... If not, I'll merge this soon.

site/index.html
kwalrath: Sorry, didn't see this until now. @filiph, what do you think?

kwalrath:
With my latest (last?) push, here are the descriptions:

Console Application - a command-line application sample.
Dart Package - a starting point for Dart libraries or applications.
Shelf Web Server - a web server built using the shelf package.
AngularDart Web App - a web app with material design components.
Bare-bones Web App - a web app that uses only core Dart libraries.
StageXL Web App - a starting point for 2D animation and games.
  
```
Available generators:
  console-full   - A command-line application sample.
  package-simple - A starting point for Dart libraries or applications.
  server-shelf   - A web server built using the shelf package.
  web-angular    - A web app with material design components.
  web-simple     - A web app that uses only core Dart libraries.
  web-stagexl    - A starting point for 2D animation and games.
```

PR #414, opened 2017-05-22
rename 'Shelf Web Server' to 'Web Server'

devoncarew:
Rename the `Shelf Web Server` app to `Web Server`. It's the only server app in the template list now, and we're not really trying to promote shelf (we still mention shelf in the template description).

### DIFF
--- a/lib/generators/server_shelf.dart
+++ b/lib/generators/server_shelf.dart
@@ -7,11 +7,11 @@ import '../src/common.dart';
 import 'server_shelf_data.dart';
 
 /**
- * A generator for a hello world command-line application.
+ * A generator for a server app built on `package:shelf`.
  */
 class ServerShelfGenerator extends DefaultGenerator {
   ServerShelfGenerator()
-      : super('server-shelf', 'Shelf Web Server',
+      : super('server-shelf', 'Web Server',
             'A web server built using the shelf package.',
             categories: const ['dart', 'shelf', 'server']) {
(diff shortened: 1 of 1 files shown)


devoncarew:
@kevmoo

kwalrath:
OK, but know that every time we change a template name, we need to redo screenshots throughout the docs. (See [docs that depend on the list of templates](https://github.com/google/stagehand/wiki/Docs-that-depend-on-stagehand#pages-that-depend-on-the-list-of-templates).) What makes it even more fun is that I edit the text in these screenshots by hand to avoid exposing usernames.

I guess what I'm saying is that if you've been wanting to make any other string changes, now is the time to do it (rather than spacing them out over multiple releases).

devoncarew:
Thanks for the reminder! (and sorry for the additional work)

devoncarew:
@lukechurch may motivate a discussion around the 'low-level web' / 'bare-bones web' template name.

lukechurch:
yeah, I need to have that discussion with @kevmoo and @filiph. I think we may also want to re-order the list.

I think we're probably ok delaying the docs refresh until after that discussion and another update. I don't think the Shelf/non-Shelf rename is likely to cause substantial confusion.

devoncarew:
the reordering is done - that'll be in the next IntelliJ release (2017.2). It's sorted lexically by template name now - AngularDart is at the top of the list

Issue #389, opened 2017-04-14
Have an initial IntelliJ launch configuration for Web Applications


jwren:

kwalrath:
Can you give some more details on what this means?

chalin:
@kevmoo - any opinion about this?

Launching a web app isn't very involved -- select index.html and then "Run...". These are the steps that we ask users to perform in our Angular docs. Is it really worth adding a launch config for this?

I'm concerned that we might then end up having to add launch configurations for all templates, for IntelliJ and potentially other editors.

Jaime: was there more to the launch config that you wanted to add?

devoncarew:
For some context on this, we have a goal for improving the getting started experience for Dart web apps in Q2, and associated with that, streamlining the run/debug experience.

We may add IntelliJ launch configs to other templates, but there are no plan to do so now.

We won't be adding launch configs for any other editors - we're only looking to streamline the IntelliJ / WebStorm experience here.

Another item we've encountered in UX reviews is there are are too many new project templates now - there's too much choice for the user and not enough clarity on which is the 'right' template. It was always a goal of stagehand to be proscriptive - to have a few high quality templates rather than supporting many different project types and configurations. I'll file a separate issue, but we'll likely need to drop a few templates and have clearer guidelines about when new ones are added.

alexander-doroshko:
Not sure what this request is about. The Dart plugin itself already creates a ready-to-use run configuration for `web/index.html` file after creating any web project using stagehand. And in case of command-line app IDE generates corresponding Dart Command Line App run configuration. Link to the corresponding code: https://github.com/JetBrains/intellij-plugins/blob/master/Dart/src/com/jetbrains/lang/dart/projectWizard/DartProjectTemplate.java#L101

If anything could be done better we can do it at Dart plugin end. I don't think stagehand has to generate IDE-specific files.

PR #416, opened 2017-05-23
Add an initial IntelliJ JavaScript Debug launch configuration for the AngularDart sample.

jwren:
@kwalrath  
@devoncarew

### DIFF
--- a/lib/generators/web_angular_data.dart
+++ b/lib/generators/web_angular_data.dart
@@ -9,6 +9,12 @@ const List<String> data = const [
 aWxkLwojIFJlbW92ZSB0aGUgZm9sbG93aW5nIHBhdHRlcm4gaWYgeW91IHdpc2ggdG8gY2hlY2sg
 aW4geW91ciBsb2NrIGZpbGUKcHVic3BlYy5sb2NrCgojIERpcmVjdG9yeSBjcmVhdGVkIGJ5IGRh
 cnRkb2MKZG9jL2FwaS8K""",
+  ".idea/runConfigurations/angular_dart.xml",
+  "text",
+  """PGNvbXBvbmVudCBuYW1lPSJQcm9qZWN0UnVuQ29uZmlndXJhdGlvbk1hbmFnZXIiPgogICAgPGNv
+bmZpZ3VyYXRpb24gbmFtZT0ibG9jYWxob3N0OjgwODAiIHR5cGU9IkphdmFzY3JpcHREZWJ1Z1R5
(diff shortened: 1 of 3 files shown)


kwalrath:
IntelliJ only, not WebStorm?

jwren:
WebStorm as well.  I say IntelliJ, but I mean any JetBrains IDE.

kwalrath:
The same problem occurs in IntelliJ 2017.1.

For either one, if I move the .idea directory elsewhere, I can open the project successfully. (I can then move it back, and the IDE will behave.)

alexander-doroshko:
Please see my comment in https://github.com/google/stagehand/issues/389#issuecomment-294358780
I think stagehand shouldn't generate IDE-specific files for the following reasons:
- General consideration: such a tool should not be tied to any specific IDE. It's great when the toolchain (stagehand, VM, pub, etc.) allow playing with the technology using just a command line, without using the IDE at all. And at the moment it is the case.
- Run configuration is **already** created by the IDE right after passing new Dart project wizard and calling stagehand.  See [the code](https://github.com/JetBrains/intellij-plugins/blob/master/Dart/src/com/jetbrains/lang/dart/projectWizard/DartProjectTemplate.java#L101). Are there any known bugs or usability issues with it? Let's fix them at the IDE end!
- IDE always knows how to create the best possible run configuration, depending on the IDE version or whatever else. Stagehand can't know it. Each IDE version is compatible with itself by definition.
- And one more show-stopper. Stagehand has no clue what URL to put into the generated run configuration. Why `http://localhost:8080`? No server is there with this port. IDE's default is `63342` and the correct URL is like `http://localhost:63342/project_name/web/index.html`. When opened - IDE starts Pub Serve on a random free port and redirects there, Pub's port is not known in advance. But even the `63342` port must not be hardcoded as there are use cases when users change it in the IDE settings. So Stagehand can't know the correct URL. But the IDE knows everything it needs.

alexander-doroshko:
And one more stopper:
- IntelliJ IDEA allows having more than one Dart modules. In an existing project, you may click File -> New module and create one more Dart project from stagehand template as a new module of the same IntelliJ IDEA project. In this case new module content folder location and existing `.idea` folder location are not dependent, they may be completely in different places on the computer. Stagehand knows neither where to generate `.idea/runConfiguration/angular_dart.xml` file, nor whether it is going to introduce conflicts with existing RCs with the same name.

alexander-doroshko:
And one more:
- IntelliJ IDEA Community Edition doesn't have JS Debugging support. Generating a run configuration there causes a confusing error message. Stagehand doesn't know whether the client supports JS debugger or not.

kwalrath:
Closing due to issues raised by @alexander-doroshko.

PR #420, opened 2017-06-01
Add commented out pubspec directions to enable DDC mode for pub serve

jwren:
@kevmoo @kwalrath


templates/web-angular/pubspec.yaml
```diff
@@ -25,3 +25,8 @@ transformers:
 - test/pub_serve:
     $include: test/**_test.dart
 - dart_to_js_script_rewriter
+
+# Enable these settings to have pub serve use DDC.
```
kevmoo: I'd block on doing this until:

- 1.24 is out
- We have content on the site we can link folks to.

Also. We should call it 'dartdevc' – DDC is really an internal name

kwalrath: I'll work on that content now...

jwren: Sounds good to me, updated to be "dartdevc".

Land whenever you see fit.

templates/web-angular/pubspec.yaml
```diff
@@ -25,3 +25,8 @@ transformers:
 - test/pub_serve:
     $include: test/**_test.dart
 - dart_to_js_script_rewriter
+
+# Enable these settings to have pub serve use dartdevc.
+#web:
+#  compiler:
+#    debug: dartdevc
```
devoncarew: nit: I'd add a trailing newline here (and in the other pubspec)

jwren: Done

chalin:
Two points that might need explaining (though I understand that this might be part of what @kwalrath is currently documenting):
- Why is `dartdevc` a value of the `debug` parameter? It might seem more comforting if it were set directly via `compiler: dartdevc` :)
- Are there any interdependencies between `dart_to_js_script_rewriter` and the ddc-based compiler transforms? E.g., must one transformation be invoked before the other or is order irrelevant?

templates/web-simple/pubspec.yaml
```diff
@@ -16,3 +16,8 @@ dev_dependencies:
 
 transformers:
 - dart_to_js_script_rewriter
+
+# Enable these settings to have pub serve use dartdevc.
```
kwalrath: Should we mention something about 1.24 here?

Maybe:

Enable these settings in 1.24+ to make pub serve use dartdevc.

templates/web-angular/pubspec.yaml
kwalrath: Content is up at https://webdev.dartlang.org/tools/dartdevc, but I think we still need to mention 1.24 in the comment. (I can make that change, if you like @jwren.)

jwren: Yes, thank you @kwalrath

kwalrath:
Good to go, @kevmoo? I'll do a separate PR for the comment, and I don't plan to release the new stagehand until 1.24 goes out.

templates/web-simple/pubspec.yaml
kevmoo: When using Dart SDK 1.24 or newer.

Also: should mention this only affects `serve` by default – since `pub build` isn't run in debug mode (unless you opt in)

Issue #425, opened 2017-06-08
Update example angular app to handle "enter" key to submit item


kevmoo:
If it's not too complex – such a great user experience

PR #426, opened 2017-06-08
Add `enter` key support to web-angular template

kevmoo:
A better example of proper usage, too – IMHO

Closes https://github.com/google/stagehand/issues/425


kwalrath:
I've wanted this for a long time! @chalin, what do you think?

Issue #422, opened 2017-06-02
Dev build is broken


kwalrath:
We haven't had a build succeed for 17 days (https://travis-ci.org/google/stagehand/jobs/233024878).

@kevmoo or @devoncarew can one of you help?

kevmoo:
Confirmed I get the same behavior on my local machine – clearly a new issue
w/ Dart 1.24

This is failing with the angular_test stuff

matanlurey:
It looks like a timeout from pub serve. I don't know specifically why unfortunately.

devoncarew:
This:

```
00:00 +0: loading test/app_test.dart
00:01 +0: loading test/app_test.dart
00:02 +0: loading test/app_test.dart
00:03 +0: loading test/app_test.dart
00:04 +0: loading test/app_test.dart
00:05 +0: loading test/app_test.dart
00:06 +0: loading test/app_test.dart
00:07 +0: loading test/app_test.dart
00:08 +0: loading test/app_test.dart
00:09 +0: loading test/app_test.dart
00:10 +0: loading test/app_test.dart
00:11 +0: loading test/app_test.dart
00:12 +0: loading test/app_test.dart
00:13 +0: loading test/app_test.dart
00:14 +0: loading test/app_test.dart
00:15 +0: loading test/app_test.dart
00:16 +0: loading test/app_test.dart
00:17 +0: loading test/app_test.dart
00:18 +0: loading test/app_test.dart
00:19 +0: loading test/app_test.dart
00:20 +0: loading test/app_test.dart
00:21 +0: loading test/app_test.dart
00:22 +0: loading test/app_test.dart
00:23 +0: loading test/app_test.dart
00:24 +0: loading test/app_test.dart
00:25 +0: loading test/app_test.dart
00:26 +0: loading test/app_test.dart
00:27 +0: loading test/app_test.dart
00:28 +0: loading test/app_test.dart
00:29 +0: loading test/app_test.dart
00:30 +0: loading test/app_test.dart
00:31 +0: loading test/app_test.dart
00:32 +0: loading test/app_test.dart
00:33 +0: loading test/app_test.dart
00:34 +0: loading test/app_test.dart
00:35 +0: loading test/app_test.dart
00:36 +0: loading test/app_test.dart
00:37 +0: loading test/app_test.dart
00:38 +0: loading test/app_test.dart
00:39 +0: loading test/app_test.dart
00:40 +0: loading test/app_test.dart
00:41 +0: loading test/app_test.dart
[...]

PR #429, opened 2017-06-12
Remove allow_failures for dev channel

chalin:
See https://github.com/google/stagehand/issues/422#issuecomment-307879808.

cc @kwalrath


kwalrath:
I'm going to wait to merge this until after we ship the new stagehand version. Just in case. :)

PR #431, opened 2017-06-12
Update stagehand version to 1.1.4

kwalrath:
@devoncarew I'm curious why we have APP_VERSION in cli_app.dart... The tests pass even if it's not updated.


devoncarew:
lgtm

> I'm curious why we have APP_VERSION in cli_app.dart

So that stagehand can respond to `--version` :)

> The tests pass even if it's not updated.

Agree, it would be good to have a test fail if the pubspec version != the cli_app.dart version.

kwalrath:
Thanks for the explanation. I'll file a bug.

PR #441, opened 2017-08-31
Ng 4

kevmoo:


templates/web-angular/pubspec.yaml
```diff
@@ -8,20 +8,20 @@ environment:
   sdk: '>=1.23.0 <2.0.0'
```
chalin: Probably want 1.24 as lower bound.

kevmoo:
Yup!

chalin:
The rest LGTM (though you might need to regenerate).

kevmoo:
Oops...yup

kevmoo:
@kwalrath @matanlurey – I raw w/ the LGTM from @chalin 

I'd like to get this out ASAP.

kwalrath:
@matanlurey: Will angular 4.0.0 work with 1.24? I'm asking because I'm not sure whether we provide kiosks at developer events, but if we do, they probably still have 1.23 (like they did for google i/o).

PR #444, opened 2017-09-01
Enable many lints, fix SDK constraint

kevmoo:


devoncarew:
Looks like there's a CI failure: https://travis-ci.org/dart-lang/stagehand/jobs/278833848#L703

pq:
> I'm curious where you're copying the lints options from, and if we should have a canonical recommended list for dart-lang/ projects.

👍 

cc @filiph who has thought about this a bit too.

FWIW this [file](https://github.com/dart-lang/linter/blob/master/example/all.yaml) contains all the lints (and is sure to since there's a test that will fail the build if it doesn't).  Our thinking was that this directory (or some place similar) would be a reasonable place to stash canonical lint sets.

kevmoo:
@pq @devoncarew PTAL

tool/grind.dart
```diff
@@ -63,41 +63,41 @@ test() => new TestRunner().testAsync(files: 'test/validate_templates.dart');
...
 }
 
 Iterable<String> _traverse(Directory dir, String root) sync* {
   var files = _listSync(dir, recursive: false, followLinks: false);
-  for (FileSystemEntity entity in files) {
+  for (var entity in files) {
     if (entity is Link) continue;
 
-    String name = path.basename(entity.path);
+    var name = path.basename(entity.path);
```
devoncarew: :violin: oh, the types, I will miss them

devoncarew:
lgtm

tool/grind.dart
kevmoo: inference is great...there's a whole team in Portland working on it 😄

devoncarew: Yeah, but the inferred type info will never show in casual tools, like in code review on github.

PR #440, opened 2017-08-30
Avoid horizontal scroll in web-simple

a14n:

PR #446, opened 2017-09-14
Refactor and enhance Travis checks

### LINKED ISSUE #424: chore(travis): split app tests from the rest of the build
@kevmoo's request from https://github.com/google/stagehand/pull/423#issuecomment-305897691:
> I'd rather split out the matrix so the angular test stuff is it's own test.
Then we can mark *that* as failing.

cc @kwalrath

### LINKED ISSUE #432: Enhance Travis checks: add version test, etc.
Our tests should fail if APP_VERSION in `lib/src/cli_app.dart` is different from the version in `pubspec.yaml`.

chalin:
- Move web-angular test into its own job. Fixes #424.
- Add an app version consistency check. Fixes #432.
- Add a check that grind was run, as mentioned in #440.
  <br>**Alternative:** instead we could just run grind `build` before deploying and avoid committing the built/generated files.

Note that `tool/version-check-and-update.sh` can be run to update the `lib/src/cli_app.dart` file using the app version in `pubspec.yaml`.

cc @kwalrath

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,31 @@ sudo: false
 dart:
   - stable
   - dev
-## The dev channel is stable enough, report failures now.
-# env: # important: w/o this top-level env, no job will be allowed to fail.
-# matrix:
+
+env: # important: w/o this top-level env, no job will be allowed to fail.
+  global:
+    - DISPLAY=:99.0
+  matrix:
+    - CI_JOB=tool/travis.sh
+    - CI_JOB=tool/version-check-and-update.sh
+    - CI_JOB=tool/build-check.sh
+    - CI_JOB=tool/web-angular-test.sh
+
+matrix:
+  exclude:
+  - dart: dev
+    env: CI_JOB=tool/version-check-and-update.sh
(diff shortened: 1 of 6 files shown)


kevmoo:
LGTM, but I defer to @devoncarew

.travis.yml
```diff
@@ -3,16 +3,31 @@ sudo: false
...
   - dev
-## The dev channel is stable enough, report failures now.
-# env: # important: w/o this top-level env, no job will be allowed to fail.
-# matrix:
+
+env: # important: w/o this top-level env, no job will be allowed to fail.
+  global:
+    - DISPLAY=:99.0
+  matrix:
+    - CI_JOB=tool/travis.sh
```
devoncarew: I would have pushed these checks into the grind script, so that more of the CI is implemented in dart, instead of being partly implemented in bash and dart. Not a blocker for this PR, but my 0.02 :)

chalin: Ok. Will explore that the next time 'round.

PR #448, opened 2017-09-25
Remove -dev.infinity from SDK constraint

kevmoo:


kevmoo:
This is actually the **reason** I started the last PR – ha!

PR #450, opened 2017-10-03
Fixed: license text is word complete

agamdua:
Hey [again](https://github.com/dart-lang/site-webdev/pull/1012) Dart team folks!

I noticed while generating a new angular project that the license text seems to have been taken from a template, but is not complete.

I did a bit of research on the interwebs and it shows that the language is either as shown in the diff below, or is the name of the author. 

That leaves us with two main tasks:

 - [ ] agree upon whether to use this text or `__author__` (its a [valid regex](https://github.com/dart-lang/stagehand/blob/master/lib/src/common.dart#L13) that is used in the [first line](https://github.com/dart-lang/stagehand/blob/master/templates/web-angular/LICENSE#L1))
 - [ ] update the other licenses according to the above decision

```
 $ git grep "<COPYRIGHT HOLDER>"
templates/console-full/LICENSE:DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
templates/package-simple/LICENSE:DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
templates/server-shelf/LICENSE:DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
templates/web-simple/LICENSE:DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
templates/web-stagexl/LICENSE:DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
```

### DIFF
--- a/templates/web-angular/LICENSE
+++ b/templates/web-angular/LICENSE
@@ -3,6 +3,7 @@ All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
+
     * Redistributions of source code must retain the above copyright
       notice, this list of conditions and the following disclaimer.
     * Redistributions in binary form must reproduce the above copyright
@@ -15,7 +16,7 @@ modification, are permitted provided that the following conditions are met:
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY
 DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
 (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
 LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND


agamdua:
It would appear that this test failure reflects the diff, which is a good thing.

I will figure out the test updates if the team conditionally accepts the patch.

kwalrath:
Thanks for the research, @agamdua. I'll look into this.

agamdua:
Awesome and thanks for taking the time to respond to each of my pull requests / issues!

kwalrath:
Thanks for taking the time to write the PRs/issues! It's easy for us to miss or forget about these, so it really helps when people let us know about problems, whether they're big or small.

kwalrath:
Hi, I haven't forgotten about this and am still looking into it...

kwalrath:
I have the solution! It turns out that the license at the top of the repo is enough, so all we need to do is add a link to it.

Could you please:

* **Remove** the template LICENSE files.
* **Delete** the copyright lines at the top of each source file. (They don't make sense if the templates don't have LICENSE files.)
* **Add** a link from each template's README to the top-level LICENSE file: https://github.com/dart-lang/stagehand/blob/master/LICENSE

The link in the READMEs should be something like this:

```
See the [LICENSE](https://github.com/dart-lang/stagehand/blob/master/LICENSE).
```

kwalrath:
@agamdua can you make this change, or should I?

agamdua:
@kwalrath 👋 

I like your idea of one template.

I do disagree that this is the right way to change things, the reasons being:

1. the impact of this change would be that every project generated with stagehand will have this at the top of the license:

```
Copyright 2014, the Dart project authors. 
```
[src](https://github.com/dart-lang/stagehand/blob/master/LICENSE#L1)

I don't think that's the intention of the project at all.

2. I am also not aware of legal implications of not actually packaging a license with the source code and have the "docs" simply _refer_ to it. I would prefer that the code generation tool would tell me explicitly to "bring my own license" instead (I don't actually want this, but if we are removing the license from the generated repository that would be a better way to go).

3. Not pinning a version to the license can also result in all sorts of problems (referring to the link to `master`).

I would rather think that `stagehand` should provide a default license file which makes the correct attributions or does not provide one at all - that also seems to be the intention of the project so far, sans the little 🐛 we need to squash.

One way to do this is to change every template in every project template, another is to link to a common file that can be accessed during stagehand use. I am not sure how to go about the latter, but I am pretty sure I can figure it out, especially if you'd like to guide me and state that you prefer that approach.

Thanks again for taking a look!

kwalrath:
@agamdua do you mean that that line would literally be inserted into every generated project's file(s)? Because that isn't the intent.

kwalrath:
Since this has morphed so much since your original commit, I'm going to close this and handle the boring details myself.

Thanks very much for kicking off this discussion, @agamdua!

agamdua:
I do not think it has changed at all - it would be nice if you mentioned your thought process, but you're not obliged to.

I think there's a communication gap and I get a sense of expectation that I need to respond sooner than I am able to which is making it harder to really do anything in the way of "collaboration". This is something I'm doing on my free time, and I get the sense that if the contribution is to be attributed to me I need to respond during work hours, even if they are non-_critical_ to the functioning of the project.

This is also the second of 3 instances of Dart related projects where I've received a "I'll just do it myself" response to a pull request and I'll admit there's no obligation to landing a commit from an external contributor, it does definitely have a negative impact on my desire to contribute in the future. The contributing guidelines imply it's ok to make pull requests for small changes. I understand it's hard to have someone new contributing, I'd prefer if that would be made more explicit in the contributing guidelines.

**regarding the actual diff**

As I see it, the original commit with the original description would be perfectly reasonable as a fix. There's potentially room for improvement to use one template but that is very specifically an _enhancement_ and not part of the original fix.

To your particular question above, a project generated by stagehand should in no way contain a reference to the Dart project as the copyright owners which is which the license you have linked to does. That is a license for all projects under the Dart umbrella and not for anyone else.

kwalrath:
Sorry, I can't talk about this in public, but I'd be happy to talk with you privately. As you can imagine, license/legal issues can get tricky.

I do respect you spending your time and effort on contributions (issues, PRs, ...), and I don't expect immediate responses. It's just... licenses are tricky.

PR #451, opened 2017-10-03
Update: installation documents to be explicit about checking version

agamdua:
The docs do not specify what it means that the "Dart SDK is on your path".
For someone who knows what this is, it is redundant and for others it may not mean much.

The patch intends to make it a more actionable step, if the command here fails because the Dart SDK is not found, *nix systems will at least complain with a standard `command not found: dart` and this can be debugged independently.

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -64,8 +64,14 @@
       <h1 id="get-started">Get Started</h1>
 
       <p>Installing Stagehand requires the Dart SDK, version 1.6 or greater.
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks, @agamdua. This will go public the next time we publish a release.

agamdua:
Awesome!

kwalrath:
The fix is now visible at http://stagehand.pub. Thanks again, @agamdua!

PR #465, opened 2018-02-01
fix version checking script

chalin:


chalin:
Travis is happy. Minor fix to script, not impacting any of the templates. Merging.

PR #477, opened 2018-04-14
update to the latest args

devoncarew:
- `stagehand` specified an older args, and was causing other packages to resolve to a pre-1.0 args


devoncarew:
I didn't know about the `2-dev` branch :)

Looks like the args dep is already upgraded there.

PR #479, opened 2018-04-16
upgrades for dart 2

devoncarew:
- make sure that the template generation can work when run in `--preview-dart-2`
- remove some explicit instructions in the  `WebAngularGenerator` template to  run pub serve


kevmoo:
@devoncarew – squash merge is your friend! 😄

PR #482, opened 2018-04-18
Analytics: Create a session for each invocation, include OS and Dart version

kevmoo:


kevmoo:
> These are custom dimensions, right?

Yup! CC @chalin

PR #484, opened 2018-04-19
Template updates

kevmoo:


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ Project and template file changes:
 - Dart 2 changes:
   - Renamed variables like, `JSON`, `PI`, etc. to `json`, `pi`, etc.
 
+Server-shelf template changes:
+
+- Updated to latest dependencies.
+- Using `int.tryParse` introduced in Dart `2.0.0-dev.48`.
```
chalin: How about:
> - Using `int.tryParse()` instead of `int.parse()`

I.e.,:
- Doc guidelines suggest putting parentheses after a function/method name.
- I'd drop the mention of a specific version of Dart.

Maybe:

PR #488, opened 2018-04-19
Migrate from bootstrapStatic to runApp

kevmoo:
Closes https://github.com/dart-lang/stagehand/issues/486


templates/web-angular/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ environment:
   sdk: '>=2.0.0-dev.48.0 <2.0.0'
 
 dependencies:
-  angular: ^5.0.0-alpha
+  angular: ^5.0.0-alpha+10
```
chalin: I'd suggest leaving this as `^5.0.0-alpha`.

kevmoo: Forcing folks to get the version that contains `runApp`.

Eventually this will just be `^5.0.0` – but until then...

chalin: Ah right.

PR #500, opened 2018-04-30
Prepare for v2 release

kevmoo:
Closes https://github.com/dart-lang/stagehand/issues/470


pubspec.yaml
```diff
@@ -7,12 +7,12 @@ name: stagehand
 description: >
   A scaffolding generator for your Dart projects. Stagehand helps you get set
   up!
-version: 2.0.0-alpha-dev
+version: 2.0.0
```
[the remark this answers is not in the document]
kevmoo: We have to – or else `pub global activate` won't get it.

No one depends on stagehand as a package. The version doesn't matter much AFAIC

kevmoo:
Will wait on a 👍 from @kwalrath before I land/publish

lib/src/cli_app.dart
```diff
@@ -17,7 +17,7 @@ import 'package:usage/usage_io.dart';
 const String appName = 'stagehand';
 
 // This version must be updated in tandem with the pubspec version.
-const String appVersion = '2.0.0-alpha-dev';
+const String appVersion = '2.0.0';
```
chalin: Btw, you don't need to update this manually anymore, the `version-check-and-update.sh` script does that for you when you run it. /cc @kwalrath

kevmoo:
Cool!

kwalrath:
@kevmoo let me make sure I understand: We can merge this, with no change to IDE behavior, but as soon as we `pub publish` it, EVERY IDE that uses stagehand will switch (on restart) to Dart 2 templates. So publishing this essentially makes Dart 2 the default everywhere, and we'll have to update all of our codelabs and websites accordingly.

Here's what I think we'll need to coordinate:

* https://webdev.dartlang.org will simultaneously switch to the v5 content (currently at https://webdev-dartlang-org-dev.firebaseapp.com).
* We'll need to update (with warnings, at the very least) or take down these codelabs:
  * https://codelabs.developers.google.com/codelabs/your-first-angulardart-web-app
  * https://codelabs.developers.google.com/codelabs/angulardart-firebase-web-app
  * https://webdev.dartlang.org/codelabs/angular_components

Does that sound right? Are we ready for the webdev switch?

/cc @matanlurey

kevmoo:
> as soon as we pub publish it, EVERY IDE that uses stagehand will switch (on restart) to Dart 2 templates

Nope. Only IDEs on machines that have at least Dart 2.0.0-dev.51.

Everyone else will get the v1 flavor. Since the web templates in v1 assume you have pub serve, this just makes sure folks using dev/be SDKs aren't thrown into a mess.

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 2.0.0-alpha-dev
+## 2.0.0
 
 Project and template file changes:
 - In `pubspec.yaml` files:
```
kwalrath: update minimum sdk

kwalrath: Also remove the item about enabling the analyzer plugin?

kevmoo: The analyzer plugin is still enabled – just you don't need the `true` bit any more to do it.

PR #501, opened 2018-04-30
chore: better release version cross check

chalin:


chalin:
@kevmoo - this will now catch issues like https://github.com/dart-lang/stagehand/pull/500#issuecomment-385412787.

PR #502, opened 2018-04-30
Remove angular: enabled: true, use array syntax

MichaelRFairhurst:
No SDK restriction for this.

The only quirk is that this won't work for those on 0.0.14 of the plugin (or newer) without at upgrade. If they are on dev-46 or newer, it will upgrade for them automatically, otherwise they'll have to upgrade manually.


kevmoo:
Update commit message to include

```
Fixes https://github.com/dart-lang/stagehand/issues/492
```

😁

kwalrath:
The wiki has instructions... somewhere.

chalin:
Or you can just run `./tool/build-check.sh` (which essentially just runs grind and complains if any of the generated files have changed).

MichaelRFairhurst:
ok now seems all in order!

PR #510, opened 2018-05-18
Support latest angular packages, bump min sdk, prepare for release

kevmoo:
Fixes https://github.com/dart-lang/stagehand/issues/508


kwalrath:
When I tried to create a web-angular project, I got lots of errors (mostly DirectiveVisitor, it looks like):

```
...
Please report a bug: https://github.com/dart-lang/angular/issues/new
'package:angular/src/source_gen/template_compiler/find_components.dart': malformed type: line 562 pos 9: cannot resolve class 'DirectiveVisitor' from '_ComponentVisitor@209334235'
    new DirectiveVisitor(
        ^


[SEVERE] build_web_compilers|entrypoint on test/app_test.dart:
Unable to find modules for some sources, this is usually the result of either a
bad import, a missing dependency in a package (or possibly a dev_dependency
needs to move to a real dependency), or a build failure (if importing a
generated file).

Please check the following imports:

`import 'package:stagehand_test_5_18/app_component.template.dart' as ng;` from stagehand_test_5_18|test/app_test.dart at 5:1

[SEVERE] build_web_compilers|entrypoint on test/app_test.dart.browser_test.dart:
Unable to find modules for some sources, this is usually the result of either a
bad import, a missing dependency in a package (or possibly a dev_dependency
needs to move to a real dependency), or a build failure (if importing a
generated file).

Please check the following imports:

`import 'package:stagehand_test_5_18/app_component.template.dart' as ng;` from stagehand_test_5_18|test/app_test.dart at 5:1

[INFO] Running build completed, took 20.7s
[INFO] Caching finalized dependency graph completed, took 302ms
[INFO] Succeeded after 21.1s with 1253 outputs (5004 actions)
Serving `web` on http://localhost:8080
Serving `test` on http://localhost:8081
[INFO] Builds finished. Safe to exit
stagehand_test_5_18$ 
```

kwalrath:
Oh wait, I was using the old version. Testing again...

kevmoo:
It worked for me. It's actually a bit urgent to get this out. Our constraints let folks get out of sync. This fixes that for the next time, too...

templates/web-angular/pubspec.yaml
```diff
@@ -5,11 +5,11 @@ description: A web app that uses AngularDart Components
 # author: __author__ <email@example.com>
 
 environment:
-  sdk: '>=2.0.0-dev.51.0 <2.0.0'
+  sdk: '>=2.0.0-dev.52.0 <2.0.0'
 
 dependencies:
-  angular: ^5.0.0-alpha+11
-  angular_components: ^0.9.0-alpha+11
+  angular: 5.0.0-alpha+12
```
chalin: IMHO, we should have kept the leading `^`, otherwise this means that we'll need to release a new version of stagehand every time a new version of angular or angular_components comes out.

templates/web-angular/pubspec.yaml
```diff
@@ -19,6 +19,5 @@ dev_dependencies:
   test: ^0.12.30
 
 dependency_overrides:
-  # Needed to ensure 0.4.0-alpha+12 (or greater) is not accessed
   # https://github.com/dart-lang/stagehand/issues/506
-  angular_compiler: 0.4.0-alpha+11
+  angular_compiler: 0.4.0-alpha+12
```
chalin: I don't think that we need the dependency_override anymore (now that we've moved to angular 5.0.0-alpha+12).

templates/web-angular/pubspec.yaml
kevmoo: Until we figure out how to not break folks, let's leave it here.

We have ideas, but for now better folks not get a broken experience and put the burden on us to release.

templates/web-angular/pubspec.yaml
kevmoo: We do because we haven't published a version of pkg:angular that properly constraints angular_compiler

templates/web-angular/pubspec.yaml
chalin: Good strategy.

templates/web-angular/pubspec.yaml
chalin: Ok, thanks for the update.

PR #512, opened 2018-05-18
add missing pkg:glob dependency

kevmoo:


pubspec.yaml
```diff
@@ -28,6 +28,7 @@ dev_dependencies:
   build: ^0.12.6
   build_config: ^0.2.1
   build_runner: ^0.8.9
+  glob: ^1.1.5
   grinder: ^0.8.0
```
chalin: We can probably drop `grinder` now, right?

kevmoo: It's used somehow with testing. If you can figure out how, I'd support it @chalin !

Issue #522, opened 2018-05-29
Fresh AngualrDart project fails to build.


kdxvc:
[**Edit** by @chalin - trimmed down the log since we don't need the whole thing to diagnose the problem]

When starting a new AngularDart project via stagehand project template, new project fails to build with errors.

C:\Users\its_me>dart --version
Dart VM version: 2.0.0-dev.58.0 (Wed May 23 20:44:51 2018 +0200) on "windows_x64"

C:\Users\its_me>pub --version
Pub 2.0.0-dev.58.0

intellij dart plugin v182.2371.1

How to reproduce.

1) open phpstorm
2) File > New Project...
3) select Dart > Generate sample content > AngularDart Web App > Create
4) new project opens in IDE
5) run pub get
```
>pub get
Resolving dependencies...
Overriding the upper bound Dart SDK constraint to <=2.0.0-dev.58.0 for the following packages:

archive, args, bazel_worker, boolean_selector, build_config, charcode, cli_util, convert, fixnum, glob, graphs, html, http, http_multi_server, http_parser, intl, io, js, logging, meta, multi_server_socket, node_preamble, package_config, path, perf_api, plugin,
protobuf, pub_semver, quiver, quiver_hashcode, sass, shelf_packages_handler, source_map_stack_trace, source_maps, source_span, stack_trace, stream_channel, string_scanner, term_glyph, tuple, typed_data

To disable this you can set the PUB_ALLOW_PRERELEASE_SDK system environment variable to `false`, or you can silence this message by setting it to `quiet`.
+ analyzer 0.32.0
+ angular 5.0.0-alpha+12 (5.0.0-alpha+13 available)
+ angular_ast 0.5.3+1
+ angular_components 0.9.0-alpha+12 (0.9.0-alpha+13 available)
+ angular_forms 2.0.0-alpha+4 (2.0.0-alpha+5 available)
+ angular_test 2.0.0-alpha+10 (2.0.0-alpha+11 available)
+ archive 1.0.33 (2.0.0 available)
+ args 1.4.3
+ async 2.0.7
+ bazel_worker 0.1.9 (0.1.10 available)
+ boolean_selector 1.0.3
+ build 0.12.6
+ build_config 0.2.6+2
+ build_modules 0.2.2+5
+ build_resolvers 0.2.0+2
+ build_runner 0.8.9
+ build_test 0.10.2+3
+ build_web_compilers 0.4.0+1
+ built_collection 3.1.1
[...]

Issue #521, opened 2018-05-25
web-angular: upgrade NG and ACX to +13, set min SDK to 2.0.0-dev.55.0


chalin:
Also address #506

chalin:
Closed by #525

PR #525, opened 2018-05-29
Upgrade min SDK and web-angular

### LINKED ISSUE #506: Remove pinning of angular_compiler
- [ ] Roll angular to a version w/ good pinning (https://github.com/dart-lang/angular/issues/1325)
- [ ] Get a version of angular components that is compatible with it

chalin:
Fixes #506, #521, #522


templates/web-angular/pubspec.yaml
```diff
@@ -5,20 +5,20 @@ description: A web app that uses AngularDart Components
...
-  angular: 5.0.0-alpha+12
-  angular_components: 0.9.0-alpha+12
+  angular: ^5.0.0-alpha+13
+  angular_components: ^0.9.0-alpha+13
 
 dev_dependencies:
-  angular_test: ^2.0.0-alpha
-  build_runner: ^0.8.7
-  build_test: ^0.10.0
+  angular_test: ^2.0.0-alpha+11
```
kevmoo: should this sync to +13, too?

templates/web-angular/test/app_test.dart
```diff
@@ -25,4 +25,6 @@ void main() {
   test('heading', () {
     expect(appPO.title, contains('My First AngularDart App'));
   });
+
+  // Testing info: https://webdev.dartlang.org/angular/guide/testing
```
kevmoo: Put this on main? or the top of the file?

More text? "To learn more about testing angular apps, seee..."

chalin: This is what we had before (and is something that @kwalrath had approved I believe. It is in the same vein as when we write simply: "More information: ..." (used in at least a dozen places on the webdev docs).

Shall we keep this for now and revisit later as needed?

kevmoo: Can leave for now. I'd be nice if this was a bit more conspicuous so we help users get started.

templates/web-angular/pubspec.yaml
kevmoo: ack

kevmoo:
@chalin – ship it!

chalin:
@kevmoo - will you be publishing a new version? If so, have you decided if you really want to bump up the version to 3.0.0 (or actually bend the semver rules for this case and keep it at 2.1.0)?

kevmoo:
> have you decided if you really want to bump up the version to 3.0.0 (or actually bend the semver rules for this case and keep it at 2.1.0)?

Let's keep w/ semver. Most users don't think about the version of this anyway. They just run 'pub global activate' (or IDEA does for them)

chalin:
> Let's keep w/ semver

SGTM

I'm not a stagehand uploader. Will you be publishing or do you want to wait?

kevmoo:
I can. Although look at the issue I just filed...

PR #527, opened 2018-05-31
web-angular: unpin sass package

### LINKED ISSUE #524: web-angular: Remove pinning of sass package
Pinning introduced by #525.

From https://github.com/dart-lang/angular_components_example/pull/84#issuecomment-392888730:
> I filed https://github.com/sass/dart-sass/issues/334. It might be fixed by the sass package. I'm also close to releasing an update for sass_builder that will effectively avoid the issue. Not sure which one will happen first.

Blocked on either:

- https://github.com/sass/dart-sass/issues/334
- Or the next release of https://github.com/dart-league/sass_builder

cc @nshahan

chalin:
Also have analyzer ignore .dart_tool and build folder contents.

Closes #524 

cc @nshahan


templates/web-angular/analysis_options.yaml
```diff
@@ -1,4 +1,5 @@
 analyzer:
+  exclude: [.*/**, build/**] # https://github.com/dart-lang/sdk/issues/33112
```
kevmoo: By default, the build directory doesn't have dart files though – if you're using release mode...

chalin: IMHO, there are enough scenarios that result in the creation of `build` folder potentially with `*.dart` files. E.g.: one such use case is to run tests using build_runner and also output files to `build`.

kevmoo: `.directories` are auto-ignored, right?

I've never seen those show up with the analyzer CLI –

chalin: As agreed offline, I've dropped the `.*/**` but kept `build/**`.

PR #549, opened 2018-06-28
misc: cleanup .gitignore files across templates

kevmoo:
Dart 2 does not create a .pub directory
`build/` is not created by pub – move it to a new section


kevmoo:
CC @kwalrath RE .gitignore changes – may be relevant to `site-www`

PR #569, opened 2018-08-02
Web templates: use build_runner ^0.10.0

### LINKED ISSUE #568: bump all web templates to build_runner: ^0.10.0

chalin:
Fixes #568


kwalrath:
It looks like we lost this change from the CHANGELOG, somehow. (See https://pub.dartlang.org/packages/stagehand#-changelog-tab-.) I'll add it back unless someone tells me that was intentional.

PR #573, opened 2018-08-14
misc: use the latest pkg:source_gen

kevmoo:


kevmoo:
> Maybe add a changelog entry?

I avoid such entries if there is no effect on user behavior.

PR #579, opened 2018-09-01
Migrate version logic to pkg:build_version

kevmoo:


kevmoo:
> Very nice! (build_version will be quite a useful utility builder.)

😄

PR #594, opened 2019-01-16
migrate template defaults to use package:pedantic for lints

### LINKED ISSUE #593: consider updating template to use `package:pedantic` lints
The current [list of lints](https://github.com/dart-lang/stagehand/blob/master/templates/analysis_options.yaml) is its own subset of the [complete list](https://github.com/dart-lang/linter/blob/master/example/all.yaml).  Historically this has been hand-edited, with the `pana` package [manually updated](https://github.com/dart-lang/pana/blob/master/lib/src/analysis_options.dart) to keep consistent.  Before we had a set of "standard" defaults this made sense but now that we have the evolving set of Google standard lints defined by [`package:pedantic`](https://github.com/dart-lang/pedantic) we should consider updating the template to use those.

@kevmoo @devoncarew  @mit-mit: any reason _not_ to prefer the `pedantic` rules for sample projects?

/cc @davidmorgan @bwilkerson

pq:
Migrates templates to use the `package:pedantic` lint rule set.

~I started with just one options file to keep word-smithing / bike-shedding in one place.  When we're happy with it, I'll update the others before proposing to land.~

**Ready for review.**

Additional context: https://github.com/dart-lang/linter/issues/1365.

Fixes #593.

/cc @kwalrath @kevmoo @chalin


kevmoo:
I'm on the fence about using the actual `pedantic` import, etc. I think it's a bit too much magic for brand new packages.

I'd rather we copy-paste the lints in the file w/ a link to the pedantic repo.

kevmoo:
Thanks. There is an upside to depending on pkg:pedantic, but then we have the discussion about "do we pin the version or set a range". If we pin it, then we have to explain why we pin it. If we set a range, then 'pub upgrade' may add new lint warnings, which is weird. I think leaving it to the user to add the dependency is goodness

pq:
@kevmoo: that's my thinking too.  I totally punted on pinning and that's probably OK.  Should we at least include the pedantic version we're deriving from in the comment?  Maybe something like:

```
# Google-default lint rules as defined in package:pedantic (v1.4.0).
# See: https://github.com/dart-lang/pedantic for details and the latest
# ruleset.
```

Back to pinning, maybe we could add a breadcrumb in the pubspec:

```
dev_dependencies:
#   include the latest pedantic rules 
#   pedantic: ^1.0.0
#   alternatively, pin to a specific version, if you want to ensure that newly added rules can't break us
#   pedantic: 1.4.0
```

/fyi @davidmorgan.

kevmoo:
Great idea!

pq:
Great. I updated one template:

https://github.com/dart-lang/stagehand/pull/594/files#diff-79e948527db3596d41d525bb56aae0cb
https://github.com/dart-lang/stagehand/pull/594/files#diff-067bb172fdf806e8624bc04f8885b458

If folks approve, I'll go ahead and update the rest.

Thanks!

/cc @kwalrath @chalin

chalin:
**My vote would be for @pq's original proposal: just import `pedantic`.**

In this way newbies get the default set of lints out-of-the-box; a list that is continually updated and curated by expert Dart users (mainly based, I assume on internal projects). Power users will know how to twiddle things :) to their liking, including if/when to pin a package to a version or not.

I don't like the non-[DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) suggestion of having a comment like "Google-default lint rules as defined in package:pedantic (v1.4.0)" and then duplicating the pedantic list (from one specific point in time). That is just more maintenance for us down the line.

My 2 cents :)

devoncarew:
> My vote would be for @pq's original proposal: just import pedantic.

Yeah, this does keep things nice and simple.

kevmoo:
VS having these out of sync w/ pedantic, I'll take the import. Proceed!

pq:
OK!  Updated to the first proposal and maybe ready to go?

PTAL.  Word-smithing welcome.

davidmorgan:
Looks good, thanks!

pq:
@kwalrath : any thoughts on wording (or otherwise)?

kwalrath:
Looking now...

templates/analysis_options.yaml
```diff
@@ -1,14 +1,14 @@
+# Defines a default set of lint rules enforced for
+# projects at Google. For details and rationale,
+# see: https://github.com/dart-lang/pedantic#enabled-lints.
```
kwalrath: see: -> see

pq: Fixed.  Thanks!

PR #604, opened 2019-05-02
Add categories to --machine output

DanTup:
This adds the (existing) categories to the JSON output so that editors can filter them.

LMK if I should bump the version/add to changelog here (or include this in the other PR elsewhere..).

@kevmoo @devoncarew FYI!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 3.3.0-dev
 
+- Include templates `categories` in the `--machine` output.
+
 ## 3.2.1
 
 - All templates:
--- a/lib/src/cli_app.dart
+++ b/lib/src/cli_app.dart
@@ -234,7 +234,8 @@ additional analytics to help us improve Stagehand [y/yes/no]?''');
       var m = {
         'name': generator.id,
         'label': generator.label,
-        'description': generator.description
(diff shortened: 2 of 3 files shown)


kwalrath:
@kevmoo and @devoncarew, who'd be a good reviewer for this?

devoncarew:
> @DanTup is this change visible to stagehand users in any way? If so, we should add something to the changelog.

It's not visible to users - just to IDEs (but a changelog entry couldn't hurt).

DanTup:
Yeah, it should probably have a changelog entry but I wasn't sure what to do about the version. @kevmoo shall I bump the version to 3.2 here and add changelog, then rebase my other PR on top (and we can agree that both changes will go into the same version so I can set the minimum version in VS Code)?

kevmoo:
Let's bump the version to 2.3.0-dev here – then we can roll in the other bits on Monday

devoncarew:
the version's been bumped to `3.3.0-dev` in a separate PR

DanTup:
Great, I've rebased this and added to the changelog. Please land this if you're still happy with it, and if possible rebase the other branch, so that I can fix up VS Code to use the categories and require a v3.3.

devoncarew:
landing

PR #611, opened 2019-06-04
Make server-shelf compatible with Cloud Run

domesticmouse:
[Cloud Run](https://cloud.google.com/run/) has a couple of requirements of apps that it attempts to run. The first is that they bind to the port specified in the PORT environment variable, and the second is that it listens to `0.0.0.0` instead of `localhost`. This change should make it as simple as adding a simple Dockerfile to the generated sample app to get it running on Cloud Run.


kevmoo:
re-run `pub run build_runner build` – would be good.

domesticmouse:
PTAL @kevmoo

templates/server-shelf/bin/server.dart
```diff
@@ -23,7 +23,7 @@ main(List<String> args) async {
       .addMiddleware(shelf.logRequests())
       .addHandler(_echoRequest);
 
-  var server = await io.serve(handler, 'localhost', port);
+  var server = await io.serve(handler, '0.0.0.0', port);
```
kevmoo: So. This is interesting. If we do this, the first run (on Mac at least) will do a security prompt since this is creating a listener on all interfaces.

Better to put in a doc comment telling folks how to deal w/ this?

templates/server-shelf/bin/server.dart
```diff
@@ -5,15 +5,15 @@ import 'package:shelf/shelf.dart' as shelf;
 import 'package:shelf/shelf_io.dart' as io;
 
 main(List<String> args) async {
-  var parser = ArgParser()..addOption('port', abbr: 'p', defaultsTo: '8080');
-
+  var parser = ArgParser()..addOption('port', abbr: 'p');
   var result = parser.parse(args);
 
-  var port = int.tryParse(result['port']);
+  var portStr = result['port'] ?? Platform.environment['PORT'] ?? '8080';
```
kevmoo: Perhaps add a comment here explaining why we're checking the environment `PORT`?

templates/server-shelf/bin/server.dart
domesticmouse: Reverted to `localhost`, but now with explicit instructions on how to adapt to Cloud Run

templates/server-shelf/bin/server.dart
domesticmouse: done

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 3.3.1
+
+- Added comments to `server-shelf` on how to use it with
```
kevmoo: Also support `PORT` env var

domesticmouse: Clarified Changelog.md

kevmoo:
In theory we could publish this now, although I don't know if there is any urgency.

CC @kwalrath

domesticmouse:
This change came out of a Flutter Hackathon event where I and an attendee tried to use Dart on Cloud Run to act as the server side of a Flutter app. We both burned the majority of the event trying to figure out what we were doing wrong. Please publish this ASAP to prevent developer frustration.

Thanks!

kevmoo:
Deployed as 3.3.1!

PR #618, opened 2019-07-10
Document VSCode Flutter Plugin requirements

domesticmouse:
PTAL @DanTup @kwalrath

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ language: dart
 sudo: required
 
 dart:
-  # Make sure this Dart SDK version is the same as the stable Flutter Dart SDK
-  # version to prevent errors when using Stagehand from inside VSCode's Dart
-  # and Flutter plugins.
+  # Make sure this Dart SDK version is no higher than the stable Flutter Dart
+  # SDK version to prevent errors when using Stagehand from inside VSCode's
+  # Dart and Flutter plugins. For more detail, please see:
+  # https://github.com/dart-lang/stagehand/issues/617
   - 2.4.0
   - dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ language: dart
 sudo: required
 
 dart:
+  # Make sure this Dart SDK version is the same as the stable Flutter Dart SDK
+  # version to prevent errors when using Stagehand from inside VSCode's Dart
+  # and Flutter plugins.
   - 2.4.0
   - dev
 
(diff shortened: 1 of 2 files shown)


domesticmouse:
Prevents a replay of https://github.com/dart-lang/stagehand/issues/617

.travis.yml
```diff
@@ -5,6 +5,9 @@ language: dart
 sudo: required
 
 dart:
+  # Make sure this Dart SDK version is the same as the stable Flutter Dart SDK
```
kwalrath: Does it really need to be “the same as”? What if stable Flutter is following an alpha/dev/dot release? (Maybe this won’t be an issue in future Flutter releases.)

domesticmouse: I want to always support Flutter's stable release for the stagehand binary, as it will be the last to update. Flutter's master, dev, and beta releases are all ahead of stable, with master and dev running ahead of released Dart versions. 

The templates should be allowed to stay in check with Dart releases as that will work for both the Dart samples and the Flutter Web samples working with `webdev`.

Does that sound about right @DanTup?

DanTup: I don't think it needs to be "the same as", but it should be "not higher than the Dart SDK included in the current stable Flutter release".

> The templates should be allowed to stay in check with Dart releases as that will work for both the Dart samples and the Flutter Web samples working with `webdev`.

I'm not sure what this bit means, but in VS Code when you open a Flutter Web projects, we'll also use the Flutter SDK (and therefore, the Dart version inside it). We'll use the standard Dart SDK for non-Flutter-Web projects though.

It may be worth including a link to the issue that has some description of exactly why this is needed (so if in future someone thinks they need to update it, they have the whole picture).

devoncarew: Users may not be using a dart sdk from the flutter sdk, or using VS Code; this seems like very specific advice for this template.

domesticmouse:
This is not for the templates but is for the stagehand tool itself. I broke stagehands VSCode integration by walking the version ahead of Flutter’s stable dart sdk version

.travis.yml
domesticmouse: This isn't a template file, it's a testing restriction for the stagehand command itself to make sure we don't break the VSCode Dart-Code plugin.

domesticmouse: Added link to issue and clarified the wording.

domesticmouse:
@kwalrath please merge, I no longer have access ;-)

PR #624, opened 2019-07-30
Remove xvfb from web-angular-test.sh

johnpryan:
I don't think xvfb is required to run chrome tests anymore, but here goes...


johnpryan:
I think this is good to merge. Chrome now [runs in a headless environment](https://developers.google.com/web/updates/2017/04/headless-chrome) but we can always [add xvfb](https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-to-run-tests-that-require-a-gui) to the travis configuration again if we need to.

PR #625, opened 2019-08-10
Update repository address after moving to dart-lang org

Fox32:
Just stumbled across this while generating a new project


kwalrath:
The build is broken because we need to compress the templates. https://github.com/dart-lang/stagehand/blob/master/CONTRIBUTING.md has instructions.

Can you run the build, @Fox32? If not, I can do it.

Fox32:
@kwalrath Give me a moment to try it

PR #626, opened 2019-08-12
Update repository address after moving to dart-lang org

Fox32:
I made a mistake and GitHub closed my previous PR (#625)...

@kwalrath My bad, next time I read the contribution guide carefully 😉 

I just rebuild the templates, but noticed that not only the template related to my changes are changed. This seems to be related to me working on Windows. I changed my line endings to keep the one from the repository and only the related templates get changed, but the `validate_templates.dart` still fails.

So I switched to my MacBook. Maybe a warning about Windows in the contribution guide would be nice.


kwalrath:
> I just rebuild the templates, but noticed that not only the template related to my changes are changed. This seems to be related to me working on Windows. I changed my line endings to keep the one from the repository and only the related templates get changed, but the `validate_templates.dart` still fails.
>
> So I switched to my MacBook. Maybe a warning about Windows in the contribution guide would be nice.

As you can probably tell, we don't use Windows much at Google.

Thank you for persevering! (And good thing you have a MacBook.)

kwalrath:
@Fox32 would you be willing to send us a PR for the contribution guide?

PR #628, opened 2019-08-13
Mention Windows issues in contibution guide

### LINKED ISSUE #627: build doesn't work on windows
See #626 for details.

We should at least update the contributor's guide accordingly. Another possibility is to make the build work on windows; I don't know how hard that would be.

Fox32:
Mention in the contribution guide, that you might run into trouble if you are using a Windows maschine.

Add a .gitattributes file to force unix-style line endings on Windows maschines.

Closes #627


CONTRIBUTING.md
```diff
@@ -36,6 +39,10 @@ pub run test
 dart test/validate_templates.dart
 ```
 
+The `validate_templates.dart` test currently doesn't support the different line endings
+on Windows machines and fails, so prefer using a Linux or Mac machine for contributing to this project.
```
kwalrath: Should we say this earlier?

Fox32: Yes we should, but I won't be able to do it this week

PR #629, opened 2019-08-22
update README: VS Code supports Stagehand

nbros:
With the Dart Code extension: see https://dartcode.org/releases/v2-23/


nbros:
@googlebot I signed it!

PR #631, opened 2019-09-12
Update to Dart 2.5

domesticmouse:


domesticmouse:
PTAL @DanTup, @kwalrath

DanTup:
Thanks for the ping! This LGTM 👍 

I don't think the issue we had previously will be an issue now that Flutter ships Dart SDKs with good version numbers (assuming that continues :-)). Previously Flutter had a `-dev` suffix, which meant it was "below" the stable version, but I think that's been fixed - the current Flutter stable SDK says it's 2.5.0 (I also just tested it, and Flutter stable's `pub` was happy to run Stagehand 3.3.3 with this change).

domesticmouse:
@dantup we still need to be careful if we push a new Dart version without a matching Flutter stable release, wherein we can bump the templates, but don’t uplift the tool itself. Thanks for checking in 😎

PR #633, opened 2019-09-17
Dropping Flutter Web sample.

domesticmouse:
Fixes https://github.com/dart-lang/stagehand/issues/632


domesticmouse:
PTAL @kwalrath

DanTup:
Removing this broke the **Flutter: New Web Project** in VS Code, since it was hard-wired to use this template.

I think this was a bit of an edge case (we don't usually reference templates by ID, but here we had to), but I thought it worth mentioning - the IDEs do use stagehand so changes like this could (though usually aren't) breaking.

I'll remove the command from the next release, but I won't bother hot-fixing it unless people start raising issues .

@pq FYI - I'm not sure if it works the same way in IntelliJ - if you're just showing the full list it probably doesn't matter, but just in case - FYI!

kwalrath:
Thanks for the info, @DanTup. I didn't know that any IDEs were hardwired to use individual templates! I think/hope IntelliJ just shows the full list.

DanTup:
Yeah, it's not very common for us to have something this specific (for this reason) but it was confusing to have Flutter's web projects appear under **Dart: New Project** (which uses stagehand) when there's a **Flutter: New Project** command alongside it (using `flutter create`) so we added some specific handling (in VS Code, at least). 

I actually carefully coded it so if additional templates were added with a `flutter-web` prefix, it would handle them gracefully - allowing the user to select from a list - I didn't think about it being removed though 🤦‍♂️! Though to be honest, it probably wouldn't have helped - for how we need to set the command up, there's no way we could dynamically hide the command (VS Code limitations) so the best we could've done is shown an appropriate error instead of a weird empty list until it was removed in another release.

kwalrath:
Related: #637

PR #645, opened 2019-12-11
Upgrade to Dart 2.7

domesticmouse:


domesticmouse:
PTAL @kwalrath

kwalrath:
We should probably disable the dev build, for now.

Also... changelog?

domesticmouse:
Changelog entry added. I think we can leave the dev build in, given that the build release re-ordering agreed elsewhere.

PR #651, opened 2020-02-04
add a new `console-simple` template

devoncarew:
- add a new `console-simple` template
- fix https://github.com/dart-lang/stagehand/issues/650

@kevmoo @kwalrath


devoncarew:
cc @mit-mit

templates/console-simple/.gitignore
```diff
@@ -0,0 +1,11 @@
+# Files and directories created by pub
+.dart_tool/
+.packages
+# Remove the following pattern if you wish to check in your lock file
+pubspec.lock
```
mit-mit: The recommendation is to check .lock in for apps, so I think we should remove this

templates/console-simple/.gitignore
```diff
@@ -0,0 +1,11 @@
+# Files and directories created by pub
+.dart_tool/
+.packages
```
mit-mit: Is this still relevant?

templates/console-simple/.gitignore
devoncarew: I think that really applies to bigger apps - more people working on it, a mature release process, ...

The .gitiignore files are the same for all templates in stagehand (and verified the same by the CI); if we want to update them we should do it in a separate PR.

templates/console-simple/.gitignore
devoncarew: `.packages` yes, but `build/` may not longer be?

kwalrath:
I wonder if we should change the README for console-full, since it differs from console-simple's by a single character.

devoncarew:
> I wonder if we should change the README for console-full, since it differs from console-simple's by a single character.

Makes sense; I updated the readme for the console-full sample to decribe a bit more what it does and help differentiate it from console-simple.

templates/console-simple/.gitignore
mit-mit: I don't understand. Why would this not be good for any app? How does the number of people matter?

templates/console-simple/.gitignore
mit-mit: Looks we we need to keep both: .dart_tool is where package_config.json goes. Closing this thread again.

templates/console-simple/.gitignore
devoncarew: I've never checked in the lock file for anything other than larger, mature apps where we need to ship reproducible builds to users. For everything else I've relied on the version ranges in the pubspec file.

I think checking in a lock file for a casual, small cli app is overkill.

Again though, the .gitignore file is standardized across all the project templates in stagehand, so if we makes changes we should change it at https://github.com/dart-lang/stagehand/blob/master/.gitignore.

devoncarew:
> LGTM! (Although I think you'll need to regenerate because I changed the README.)

Thanks! Done.

devoncarew:
@mit-mit, I'd like to land this PR (and split off discussions about changing the .gitignore files into a separate issue).

mit-mit:
Happy to track in a separate issue, but I'm inclined to immediately open a PR so make the edit then, so kindly comment over in https://github.com/dart-lang/stagehand/issues/653

PR #657, opened 2020-03-16
recommend the `any` constraint for package:test

devoncarew:
- recommend the `any` constraint for package:test

This aligns with the new guidance from the `dartdev create` command. It also means that our version recommendation won't get out of date.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.3.8-dev
+
+- Changed the dev dependency for new projects for package:test to `any`.
+
 ## 3.3.7
 
 - Added a new template: `console-simple`.
--- a/lib/src/generators/console_full.g.dart
+++ b/lib/src/generators/console_full.g.dart
@@ -53,7 +53,7 @@ bmFtZTogX19wcm9qZWN0TmFtZV9fCmRlc2NyaXB0aW9uOiBBIHNhbXBsZSBjb21tYW5kLWxpbmUg
 YXBwbGljYXRpb24uCiMgdmVyc2lvbjogMS4wLjAKIyBob21lcGFnZTogaHR0cHM6Ly93d3cuZXhh
 bXBsZS5jb20KCmVudmlyb25tZW50OgogIHNkazogJz49Mi43LjAgPDMuMC4wJwoKI2RlcGVuZGVu
 Y2llczoKIyAgcGF0aDogXjEuNi4wCgpkZXZfZGVwZW5kZW5jaWVzOgogIHBlZGFudGljOiBeMS44
-LjAKICB0ZXN0OiBeMS42LjAK''',
+LjAKICB0ZXN0OiBhbnkK''',
   'test/__projectName___test.dart',
   'text',
   '''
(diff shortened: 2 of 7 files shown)


devoncarew:
https://dart-review.googlesource.com/c/sdk/+/139680

kevmoo:
I hate having "any" anywhere.

If we do breaking changes to pkg:test, folks will break randomly and not
understand why!

mit-mit:
> If we do breaking changes to pkg:test,

If we do breaking changes, we should do them sparingly, and pre-announce them. Just like any other breaking change.

kevmoo:
> > If we do breaking changes to pkg:test,
> 
> If we do breaking changes, we should do them sparingly, and pre-announce them. Just like any other breaking change.

Absolutely.

But we should also not have folks fall into breaking changes by running `pub upgrade`. That's what you get when you have an `any` constraint

devoncarew:
cc @jakemac53 and @grouma, who were in the discussion for what to recommend to people for the `dartdev test` command.

I think we should unify what stagehand does and what `dartdev test` does. Here's the current impl. of `dartdev test`: https://github.com/dart-lang/sdk/blob/master/pkg/dartdev/lib/src/commands/test.dart#L53 - it's been moved to using `any`. Stagehand uses `^1.6.0`: https://github.com/dart-lang/stagehand/blob/master/templates/console-full/pubspec.yaml#L14.

I think I'm gravitating towards what Kevin recommends - that we don't use `any`, which would technically break the user's project if we ever released a package:test 2.0.0.

In order to not have to update the code w/ package:test releases - to reduce our maintenance burden and the chance for things to get out of date - I think we should change to recommend the last major version constraint - `^1.0.0`. This won't need updating frequently, and will resolve to whatever version solves with their other packages.

jakemac53:
This package can much more safely recommend a real version other than `any`, because it isn't a part of the SDK. We can update it at any time and people can always re-activate this package to get the newer constraint.

An SDK lives forever so we cannot imo suggest anything other than `any` there. That is the only reason I suggest `any` for that case.

grouma:
+1 to what @jakemac53 said. The likelihood that we break individuals that use `package:test` as a dev dependency is very low. Breaking changes are more for those that extend the test runner, e.g. Flutter / Google3.

devoncarew:
> because it isn't a part of the SDK. We can update it at any time and people can always re-activate this package to get the newer constraint.

Note that package:stagehand _is_ effectively part of the SDK. It's what `dartdev create` uses, pinned to a specific version of stagehand. I think we should align what stagehand and `dartdev test` recommend.

> An SDK lives forever so we cannot imo suggest anything other than any there. 

I don't understand that. We can vend a specific version of stagehand in, that recommends a specific version range of test (`^1.0.0`), and we can know that that version solves w/ the current SDK version that we're building.

jakemac53:
> Note that package:stagehand _is_ effectively part of the SDK. It's what `dartdev create` uses, pinned to a specific version of stagehand. I think we should align what stagehand and `dartdev test` recommend.

Oh - in that case then I retract my previous comment, we should use `any`.

> I don't understand that. We can vend a specific version of stagehand in (to the SDK repo), that recommends a specific version range of test (`^1.0.0`), and we can know that that version solves w/ the current SDK version that we're building.

It will give you _some_ compatible version but it could be arbitrarily old. The newer versions might very well also be compatible.

Adding a constraint on an old version is very likely to limit what versions of _other_ packages you can get (due it that old version having old version constraints on _its_ dependencies). It could even cause you to not be able to get a version solve at all.

devoncarew:
Closing as we don't want to recommend `any` for general development.

We may do it for very targeted use cases, where we can't predict which package version should solve w/ the sdk version and the user's other packages.

PR #662, opened 2020-05-15
Update .gitignores to match current guidance

mit-mit:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.3.10
```
kwalrath, line 1: Should this be part of 3.3.9? It hasn't been published yet. Of course, we still need the build to work...

/cc @devoncarew @jwren

kwalrath:
Some version questions, and it looks like the templates need to be regenerated.

CHANGELOG.md
mit-mit: I'll move it to 3.3.9

Issue #565, opened 2018-07-23
Consider adding unnecessary_new and unnecessary_const to template & stagehand lints


kwalrath:
See discussion: https://github.com/dart-lang/site-www/pull/1016#discussion_r204196792

kevmoo:
At a minimum, I think updating the templates to drop them is great.

Not 100% sure about enabling the lints for end users, though...

kwalrath:
[1.8.0 pedantic](https://github.com/dart-lang/pedantic/blob/master/lib/analysis_options.1.8.0.yaml) includes unnecessary_new & unnecessary_const, as does the [stagehand options file](https://github.com/dart-lang/stagehand/blob/master/analysis_options.yaml). When we update to 1.8.0, we should remove duplicate declarations (such as unnecessary_new & unnecessary_const) from the the stagehand options file.

kwalrath:
We should avoid having a free-floating pedantic dependency, since that's likely to break code whenever devs do a `pub upgrade`.

If we continue to use pedantic, we should have something like this in each template's `analysis_options.yaml` file:

```
include: package:pedantic/analysis_options.1.9.0.yaml
```

Currently, the analysis options files have something like this (which I'm calling a free-floating pedantic dependency):

```
include: package:pedantic/analysis_options.yaml
```

mit-mit:
I'd recommend pinning to pedantic 1.8.0 for now. We're investigating using a different default set of lints, so no reason to push additional lints right now.

PR #666, opened 2020-05-21
Update templates dependencies

adrianjagielak:
I updated templates dependencies to the latest stable versions and regenerated generated code


kwalrath:
Thanks for the PR, @adrianjagielak! I think we'll probably release 3.3.9 before merging this PR, just in case. We might also want to fix #565 for 3.3.10 (or maybe it should be 3.4.0).

/cc @kevmoo @domesticmouse @mit-mit @devoncarew

domesticmouse:
Updating deps feels like a 3.3.10, not a 3.4.0. Anyways, I'm going to land this and do a publish of stagehand, just to be a rebel. =)

Issue #672, opened 2020-07-13
Instructions say "run your app using `dart bin/__projectName__.dart`"


DanTup:
After creating an app, it says:

```
6 files written.

--> to provision required packages, run 'pub get'
--> run your app using `dart bin/__projectName__.dart`.
```

It looks a bit weird having `__projectName__` here - should it have been replaced?

devoncarew:
Yes, this is a bug. The text doesn't participate in the template replacement, so likely shouldn't have been updated as part of https://github.com/dart-lang/stagehand/pull/659.

We could do something like replace the guidance with `'dart run'`.

cc @jwren

PR #673, opened 2020-07-13
Fix for #672- update the directions in the getInstallInstructions() calls

jwren:


lib/src/generators/console_full.dart
```diff
@@ -21,5 +21,5 @@ class ConsoleFullGenerator extends DefaultGenerator {
 
   @override
   String getInstallInstructions() => '${super.getInstallInstructions()}\n'
-      'run your app using `dart ${entrypoint.path}`.';
+      'run your app using `dart run`.';
```
devoncarew, line 24: I didn't notice this on the original PR, but if you look up a few lines, I think the issue is with:

```
setEntrypoint(getFile('bin/__projectName__.dart'));
```

We'll likely need to do two changes here. The first is what you have - to change the install instructions text to not reference the `entrypoint` getter.

The other is to deprecate the `entrypoint` getter, and introduce a new API for clients to call. That might be something like `entrypointPath({@required String projectName})`, and would substitute any reference to `__projectName__` in the entry point path, similar to the logic here: https://github.com/dart-lang/stagehand/blob/master/lib/stagehand.dart#L95.

Let me know if that makes sense -

jwren:
I've looked at your comment once or twice, but I'm not sure I am clear on all the implications, etc.  Lets chat when you have a chance.

mit-mit:
Closing as not relevant as we're deprecating this for `dart create`

PR #675, opened 2020-08-05
Update to Dart 2.9

domesticmouse:
PTAL @kevmoo @mit-mit


templates/console-full/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: A sample command-line application.
 # homepage: https://www.example.com
 
 environment:
-  sdk: '>=2.8.1 <3.0.0'
+  sdk: '>=2.9.0 <3.0.0'
```
devoncarew, line 7: I'm not sure that we should follow the min / stable sdk so closely. We should likely let this be any relatively recent stable, until we start using a language feature from a newer SDK?

domesticmouse: My intent here is to make the first five minute experience for new users the best possible experience, especially once we ship null safety. My understanding is that null safety will require the latest language when it ships.

mit-mit: For an app, I see no reason why we wouldn't default to the current SDK.

domesticmouse:
I’m happy either way. I'll let @mit-mit make the final call on whether to land & publish.

/cc @kwalrath

PR #676, opened 2020-09-09
Update the minimum required version for webdev

Mixolyde:
webdev build requires this update to work with the base template


mit-mit:
Sorry for the staleness here. We're upgraded the templates to require sdk 2.12 or later, so that should be sufficient to resolve to the needed versions

PR #682, opened 2020-12-09
Pass a Uri to package:http APIs

natebosch:
Prepare for https://github.com/dart-lang/http/issues/375

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.3.12-dev
+
 ## 3.3.11
 
 - Updated minimum Dart SDK to 2.10.0 in stagehand tool and templates.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+## 3.3.12-dev
+
```
devoncarew, line 2: Perhaps `- Updates to prepare for a new version of package:http`

devoncarew (commented):
Thanks for the PR! In order to pass the CI, I believe you'll need to run `pub run build_runner build --delete-conflicting-outputs` locally, and add an entry to the changelog.

CHANGELOG.md
natebosch: FWIW we don't need to publish this now, we just need it merged to google3. I typically avoid adding changelogs that don't impact the user for the sake of having something there, but I can add something.

devoncarew: > I typically avoid adding changelogs that don't impact the user for the sake of having something there, but I can add something.

Same here, but from a cursory look at the travis failures, it looks like we have some validation that there's a changelog entry for new versions.

devoncarew (approved):

PR #684, opened 2020-12-22
Enabled and fixed a number of lints

kevmoo:


analysis_options.yaml
```diff
@@ -11,42 +11,62 @@ analyzer:
     dead_code: error
 linter:
   rules:
-    # Errors
-    - avoid_empty_else
+    - avoid_catching_errors
```
kevmoo, line 14: FYI: the deleted lints are already covered by pkg:pedantic 1.9

kevmoo:
CC @devoncarew as FYI

lib/src/common.dart
```diff
@@ -70,14 +70,14 @@ String normalizeProjectName(String name) {
 /// A key value can only be an ASCII string made up of letters: A-Z, a-z.
 /// No whitespace, numbers, or other characters are allowed.
 String substituteVars(String str, Map<String, String> vars) {
-  var nonValidKeys =
+  final nonValidKeys =
       vars.keys.where((k) => k.contains(_nonValidSubstituteRegExp)).toList();
   if (nonValidKeys.isNotEmpty) {
     throw ArgumentError('vars.keys can only contain letters.');
   }
```
parlough, line 77: I know you didn't write this, but seems appropriate to clean it up here. If not, that's fine as well.

Seems `nonValidKeys` is only used for this check, so could change this whole chunk to simply:

```dart
if (vars.keys.any((k) => k.contains(_nonValidSubstituteRegExp))) {
  throw ArgumentError('vars.keys can only contain letters');
}
```

parlough (approved):
Lots of good cleanup here 😁

lib/src/common.dart
kevmoo: solid idea!

kevmoo: done!

kwalrath (approved):
LGTM, although I'd feel better if @domesticmouse took a look too.

domesticmouse (approved):
LGTM

PR #690, opened 2021-02-17
project name must be normalized on generation

f:
TL;DR:

project name on `dart create <project-name>` must be normalized otherwise it fails while `pub get`.

## Steps:

Try to install dart project using `dart create` CLI helper.

```
$ dart create some-project
Creating /tmp/some-project using template console-simple...

  .gitignore
  CHANGELOG.md
  README.md
  analysis_options.yaml
  bin/some-project.dart
  pubspec.yaml

Running pub get...
Error on line 1, column 7 of pubspec.yaml: "name" field must be a valid Dart identifier.
  ╷
1 │ name: some-project
  │       ^^^
  ╵
```

## Expected:

It should be normalized to `some_project` but it keeps as input and throws while `pub get`.

## Actual:

Throws 

```
Running pub get...
Error on line 1, column 7 of pubspec.yaml: "name" field must be a valid Dart identifier.
  ╷
1 │ name: some-project
  │       ^^^
  ╵
```

## The issue:

Normalization is not done on neither `commands/create.dart` nor here. It should be done here anyway. (ref: https://github.com/dart-lang/sdk/blob/master/pkg/dartdev/lib/src/commands/create.dart#L95-L98)

### DIFF
--- a/lib/stagehand.dart
+++ b/lib/stagehand.dart
@@ -96,7 +96,7 @@ abstract class Generator implements Comparable<Generator> {
     Map<String, String> additionalVars,
   }) {
     final vars = {
-      'projectName': projectName,
+      'projectName': normalizeProjectName(projectName),
       'description': description,
       'year': DateTime.now().year.toString(),
       'author': '<your name>',


f:
I have no idea why `(ubuntu-latest, dev)` has not passed. I think it's not related to this PR.

mit-mit:
> /cc @mit-mit any idea why a test would be failing on Linux with Dev and Stable channels, but succeeding on beta?

We just discovered that we didn't get the beta channel rolled to 2.12.0 stable yet when we launched to stable. That will happen shortly. Then I suspect beta will start to show the same issue as stable and dev:

The CI issue seems to stem from an incompatibility in angular_components with 2.12.0 stable; I'm following on a resolution to that. cc @kevmoo 

```
The current Dart SDK version is 2.12.0.

Because angular_components >=0.13.0 depends on observable ^0.22.2 which depends on dart_internal ^0.1.1, angular_components >=0.13.0 requires dart_internal ^0.1.1.
So, because dart_internal ^0.1.1 requires SDK version >=2.0.0-dev.12.0 <2.12.0 and __projectName__ depends on angular_components ^1.0.2, version solving failed.
Error: Process completed with exit code 1.
```

mit-mit:
Merging as I believe the CI issue has been needed (and I actually need this change in a different PR).

PR #696, opened 2021-03-19
Enable null safety by-default in Dart templates

mit-mit:


mit-mit:
@natebosch can you check that these templates suggest the right package versions?

mit-mit:
Remaining CI failure seems to be an incompatibility between the angular template and 2.12, cc @kevmoo

pubspec.yaml
```diff
@@ -17,26 +17,26 @@ environment:
...
 # Add the bin/stagehand.dart script to the scripts pub installs.
 executables:
   stagehand:
 
 dependencies:
-  args: ^1.5.0
-  http: ^0.12.0
+  args: ^2.0.0
+  http: ^0.13.0
   path: ^1.7.0
```
natebosch, line 29:
```suggestion
  path: ^1.8.0
```

If we want these deps to be null safe we need to bump the minimum bound for the ones without major version changes as well.

pubspec.yaml
```diff
@@ -17,26 +17,26 @@ environment:
...
   pedantic: ^1.9.0
```
natebosch, line 30:
```suggestion
  pedantic: ^1.10.0
```

pubspec.yaml
```diff
@@ -17,26 +17,26 @@ environment:
...
-  build_runner: ^1.6.0
+  build_runner: ^1.12.0
   build_version: ^2.0.0
-  glob: ^1.1.5
-  grinder: ^0.8.0
-  source_gen: ^0.9.0
+  glob: ^2.0.0
+  grinder: ^0.9.0
+  source_gen: ^1.0.0
   test: ^1.6.0
```
natebosch, line 41:
```suggestion
  test: ^1.16.0
```

templates/console-full/pubspec.yaml
```diff
@@ -4,11 +4,11 @@ description: A sample command-line application.
...
 
 #dependencies:
-#  path: ^1.7.0
+#  path: ^1.8.0
 
 dev_dependencies:
-  pedantic: ^1.9.0
-  test: ^1.14.4
+  pedantic: ^1.10.0
+  test: ^1.16.6
```
natebosch, line 14:
```suggestion
  test: ^1.16.0
```

templates/package-simple/pubspec.yaml
```diff
@@ -4,11 +4,11 @@ description: A starting point for Dart libraries or applications.
...
 
 #dependencies:
-#  path: ^1.7.0
+#  path: ^1.8.0
 
 dev_dependencies:
-  pedantic: ^1.9.0
-  test: ^1.14.4
+  pedantic: ^1.10.0
+  test: ^1.16.6
```
natebosch, line 14:
```suggestion
  test: ^1.16.0
```

pubspec.yaml
mit-mit: Stagehand itself doesn't need to be migrated to use null safety (we're soon moving the templates to the `dart` tool and discontinuing stagehand). It just needs to create output that supports null safety.

domesticmouse (approved):

domesticmouse:
> @domesticmouse anything else we'd need to get this to `4.0.0`, not `4.0.0-dev`?

`4.0.0` SGTM

pubspec.yaml
mit-mit: Done

pubspec.yaml
mit-mit: Done

mit-mit:
From local testing, the web-angular template works with:
  1. https://github.com/google/observable/pull/103
  2. Rolling angular-components to that version (tested with an override)

cc @kevmoo fyi

pubspec.yaml
```diff
@@ -26,8 +26,8 @@ executables:
 dependencies:
   args: ^2.0.0
   http: ^0.13.0
-  path: ^1.7.0
-  pedantic: ^1.9.0
+  path: ^1.8.0
+  pedantic: ^1.10.0
```
domesticmouse, line 30: I'm curious why not ^1.11.0?

kevmoo: for pub global activated things, it doesn't really matter. There are no APIs that are being used here or causing version lock. 🤷

natebosch: `^1.10.0` allows `1.11.0`, and that is the version most folks will get anyway with this constraint.

Preemptively bumping the min version of a dependency can only introduce conflicts, not solve them. `pedantic` doesn't tend to cause conflicts - but best practice has always been to set the constraint at the lowest feature version that contains the functionality you rely on.

kevmoo:
But! This package is not really meant to be used as a dependency, so it's
not THAT big of deal.

PR #698, opened 2021-03-31
Disable null safety and prepare for beta release

mit-mit:
Prepare for a beta release of 4.0.0 (that enables null safety by default). Per @devoncarew we don't want to publish this as a stable version before the next stable SDK, as that would cause some to get it prematurely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 4.0.0
+## 4.0.0-beta.1
 
-- Enable null safety by default (min SDK constraint is 2.12)
+- Enable null safety by default (min SDK constraint is 2.12).
+  This will be released as a stable version with the next stable Dart SDK.
 
 ## 3.3.12-dev
 
--- a/bin/stagehand.dart
+++ b/bin/stagehand.dart
@@ -2,6 +2,8 @@
 // All rights reserved. Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// @dart=2.9
+
 import 'dart:async';
 import 'dart:io' as io;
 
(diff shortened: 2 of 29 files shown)


devoncarew (approved):
lgtm

Looking at this line: https://github.com/dart-lang/stagehand/blob/master/pubspec.yaml#L20 - do we want to increase the required sdk of stagehand to 2.12? That way the new version will only pub global activate into a 2.12 sdk. That's probably want we want, give that it's now creating null safe projects.

domesticmouse (approved):

domesticmouse:
I'm guessing we need to `dart migrate` the package code as part of this PR? 

Context: https://github.com/dart-lang/stagehand/pull/698/checks?check_run_id=2239836048#step:6:9

mit-mit:
I'm not sure it's worth it to migrate this to null safety when we plan on deprecating it? @devoncarew WDYT? Given it's a non-stable release is it OK to relax the constraint back down to 2.10?

devoncarew:
Ah, it looks like the checks fail now? I suspect the migration tool would make short work of this. We should get the CI back working until we do deprecate this; at a guess that might be about one stable release cycle out? The IDEs have not yet switched over, and there's some latency in when users upgrade their IDE versions.

mit-mit:
Yes, CI fails because the new 2.12 lower constraint enables nulls safety. It passes with a 2.10 lower constraint. Hence the question: Given it's a non-stable release is it OK to relax the constraint back down to 2.10?

devoncarew:
@domesticmouse - do you mind migrating package:stagehand to null safety? I think we could reasonably either migrate the code, or, update the dart file to opt-out of null safe (tag them w/ a previous language version).

mit-mit:
cc @devoncarew, CI is now passing

devoncarew (approved):

mit-mit:
@devoncarew can you please publish this? I don't have upload rights on this package.

domesticmouse:
Published

PR #699, opened 2021-04-05
Migrate web-simple template to null-safety

matsp:
Just stumpled over it while testing some ideas.


devoncarew:
I haven't reviewed this, but this null safe change wasn't already captured in https://github.com/dart-lang/stagehand/pull/696?

matsp:
Nope. Both web templates are affected. If it helps you I can also check the 'stagexl' template.

devoncarew:
Thanks! I added two reviewers; if we need these changes, we should be sure to also apply them to the copy in dart-lang/sdk.

devoncarew:
This will update the sdk version: https://dart-review.googlesource.com/c/sdk/+/194106.

mit-mit (approved):

domesticmouse (approved):

PR #704, opened 2021-05-21
Discontinue package

mit-mit:
Fixes https://github.com/dart-lang/stagehand/issues/671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 4.0.1
+
+- Added deprecation warning. See https://github.com/dart-lang/sdk/issues/46100 for details.
+
+## 4.0.0
+
+- Full release
+- Alphabetized imports
+
 ## 4.0.0-beta.1
 
 - Enable null safety by default (min SDK constraint is 2.12).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## 4.0.1
 
-- Added deprecation warning. See https://github.com/dart-lang/sdk/issues/46100 for details.
+- This package has been discontinued. Its replacement is the
+  [`dart create` command](https://dart.dev/tools/dart-tool)
+  available in all Dart SDKs since Dart 2.10.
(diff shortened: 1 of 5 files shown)


mit-mit:
cc @devoncarew @domesticmouse not entirely sure how to version this. Should it actually be 5.0.0?

devoncarew:
> cc @devoncarew @domesticmouse not entirely sure how to version this. Should it actually be 5.0.0?

Good question. My guess is that v4.0.0 is that last non-deprecated version that we want people to use, and v5.0.0 is the version where we deprecate? cc @natebosch

natebosch:
In my opinion deprecating in the current major version is sensible.

I don't know if we have prior experience adding a deprecation warning to a `pub global activate` tool.

devoncarew (approved):
👍

domesticmouse (approved):

domesticmouse:
From [the logs](https://github.com/dart-lang/stagehand/pull/704/checks?check_run_id=2639097866#step:7:7), please run `pub run build_runner build` to update the stashed version number.

mit-mit:
@domesticmouse yes, will do.

From an offline discussion with @devoncarew, I'm planning on updating this to be say that Stagehand is "Discontinued", not just deprecated. For the tools that are in the SDK, it makes sense to go through a two step process first deprecating then discontinuing, as discontinuing means the tool is no longer in the SDK. For stagehand, it will be on pub.dev marked discontinued, but you can still run the tool.

WDYT?

mit-mit:
Current build failure seems to also be happening on the master branch:
https://github.com/dart-lang/stagehand/runs/2658002053

mit-mit:
@domesticmouse @devoncarew can you please re-review as this changed substantially?

domesticmouse:
Still LGTM, no idea why the build is breaking. @kevmoo might have some insights.

mit-mit:
Given the build is already broken on master, I'm going to merge.

mit-mit:
@domesticmouse can you please publish? I don't seem to have permissions.

domesticmouse:
Published

